PR #8, opened 2023-12-04
Pr-Test

SoulMuncher:
Added .vs folder to gitignore and documented 2 bugs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ depcruise-out.dot
 depcruise-out.html
 update/
 spellmasons-mods/
+.vs/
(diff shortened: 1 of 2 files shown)


jdoleary:
What is .vs for?

SoulMuncher:
.vs/ is a directory autogenerated by Visual Studio. 

It tends to store the current state of your solution, such as which files you have open and testing breakpoints you might still have active, but I've found that it's better to keep this on your local repository instead of the shared one.

The .vs/ directory will also generate a lot of random and unnecessary whitespace/binary changes

PR #9, opened 2023-12-05
Fixed Burst Lerp

SoulMuncher:
Changed so that max damage is achieved as soon as unit circles touch, and damage linearly lerps to 0 at "out of range" distance.


SoulMuncher:
Instead of adding information about a bug in the same branch it gets fixed in, I'm going to put the details of the fix in the commit itself, and remove the bug from the docs, that way the final merge looks a little cleaner. Didn't think about it until after making the PR.

Issue #14, opened 2023-12-05
Witch Class


jdoleary:
Automatically receives Contaminate, curses cost 80% of usual cost

Balancing:
Increase radius of contaminate by 1.5x

Balance Suffocate and Poison to make them slightly more powerful

PR #27, opened 2023-12-05
Implemented basic Witch class and updated Contagion to Contaminate

SoulMuncher:
#14 

The contaminate spell was called contaminate in some places and contagion in others. I decided to synchronize most of these instances. Some localization changes will probably still be needed. The Witch also needs an image and localization entry.

In renaming contaminate files, I might have broken the links between them, and the spell does not appear to have the correct assets attached anymore.. Not sure if I need to type some command in the terminal to reinstall asset files or if the issue is elsewhere.


jdoleary:
Great work!

TODO(@SoulMuncher ):
Update 'spellIconContagious.png' in sheet1.json to get the spell icon to work again.
Also rename public/images/spell/spellIconContagious.png to spellIconContaminate.png

TODO(@jdoleary):
i18n: spell_contagious rename to spell_contaminate
i18n: class_witch
img: Add Witch upgrade image

SoulMuncher:
Updated the two things listed, image is working again

jdoleary:
Something went wrong with that last commit, sheet1.json has "26,078 additions, 10,038 deletions".  It was probably formatted during save.  My vscode config is set up to auto save files.  You can do Ctrl+Shift+P and type "File: Save without formatting".  And if that doesn't work, just revert the sheet1.json changes and I'll regenerate it from TexturePacker

SoulMuncher:
I (should) have the same config as you due to that .vscode/ folder in the repo. I did try a couple of solutions earlier, including save without formatting, but none of them worked. My guess is that the last version of the json was entirely autogenerated and that this is the first manual edit since then, meaning VS Code's autoformatting would be pretty much unavoidable for this kind of manual edit?

Either way, it looks like the easiest thing is to go ahead and revert the sheet1 change, sorry about that!

jdoleary:
Great work @SoulMuncher ! I'm excited to have this class in the game

Issue #35, opened 2023-12-06
Blessing - Send Mana


SoulMuncher:
jd
> A blessing that sends mana to another unit.  So the cost of the spell (lets say 20) will transfer that mana directly to the target(s)

SoulMuncher:
#40

PR #40, opened 2023-12-06
Added Blessing: Send_Mana

SoulMuncher:
#35

Still a few TODO's

- Proper Name
- Icon
- Description
- Sfx?
- Animation?


jdoleary:
~~When sending mana to multiple targets it still only costs 20.  This is how all spells scale with targeting.  It's just the benefit.~~
...now that I'm thinking about it, this might make a way for games with multiple players have an infinite mana hack.   Let's say instead that the mana is split between targets.

It should be in the Mana category.

SoulMuncher:
Done

PR #52, opened 2023-12-06
Quick Witch Fix

SoulMuncher:
Fixes an issue where if you choose Witch you'd also get Far Gazer's double range and half stamina.


jdoleary (approved):
Nice catch!

Issue #15, opened 2023-12-05
Refactor Timemason


jdoleary:
Timemason should get More max mana but it drains over time (rewards you to act quickly)
credit: @skillo

jdoleary:
I'm thinking double mana (max and stat mana upgrade) with losing 1% every second.  That may be too weak but I'd rather err on the side of taking too little mana vs too much

jdoleary:
I totally agree.  Go ahead and start on a stat/statmodifier system.  Looking forward to seeing what you come up with.
Remember to consider how it might interact with curses like Blood Curse or Split (both ways, adding and removing the curse)

PR #61, opened 2023-12-07
Implemented Timemason

SoulMuncher:
#15

Timemason loses 1% of max mana per second.

TODO - Update Timemason description


jdoleary:
Something is not working about the percentage.  I changed my max mana to 10000 and expected it to be 5000 by the time 50 seconds passed but it was still in the 7000s.
I need to get the update out this morning (before people wake up and get on the servers) so we'll save this one for 1.26

src/Underworld.ts
SoulMuncher, line 1: Fixed an issue where if your maxMana is greater than (100 * percent drain * your framerate), it'll drain down slower than it should, since it's limited to removing 1 mana per frame.

jdoleary (approved):

PR #85, opened 2023-12-09
In game chat

Couls:
Implement an in-game chat!

What this PR does:
Add an open chat hotkey into the settings
Allows you to open the chat when pressing it
Sends chat messages to everyone connected to the server in multiplayer
Colors your username based on your robe colors
Allows you to scroll up and down the chat
Automatically hides chat when clicking outside of it
Brings up chat temporarily when a message is received

What the system can be expanded to do later:
chat commands, either through mods or the system allowing more precise modification of the game for players
Tabs for different types of information
Logging damage and spells used, players joining and disconnecting, admin commands
Typing indicators
Timestamps
Smoother transitions and animations for the chat

image


src/style.css
```diff
@@ -2031,7 +2033,11 @@ body.auto-camera #camera-recenter-tip {
 }
 
 #chatbox {
-  display: none;
+  /* hacky but we keep it displayed as a heightless block so that the transition works display: block; */
```
Couls, line 2036: Would like comments on this method, it may not be worth it to have chat transitions at all due to the hack required to get it to fade in and out through CSS

src/graphics/ui/Chat.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (document.activeElement !== elChatinput) {
+          document.body.classList.toggle('showChat', false)
+        }
+      }, NotificationTime)
+    }
+  }
+}
+
+export function focusChat() {
+  elChatinput?.classList.toggle('disabled', true) // disable it so the first keypress doesn't go into chat
```
jdoleary, line 73: I'm still getting a `t` character in the input when I open it (I'm testing in Chrome)

src/graphics/ui/eventListeners.ts
```diff
@@ -122,185 +123,195 @@ function handleInputDown(keyCodeMapping: string | undefined, overworld: Overworl
...
-        x: mouseTarget.x,
-        y: mouseTarget.y
-      });
-      break;
-    case 'recenterCamera':
-      if (globalThis.player?.isSpawned) {
+  if (document.activeElement === Chat.elChatinput) {
+    return;
+  }
+  else {
```
jdoleary, line 129: The early `return` is sufficient, there's no need for the else block

src/style.css
jdoleary: I agree, I think it should just "pop in" (which is also what the other UI elements like the inventory do)

src/graphics/ui/Chat.ts
```diff
@@ -0,0 +1,76 @@
...
+          document.body.classList.toggle('showChat', false)
+        }
+      }, NotificationTime)
+    } else {
+      chatTimeout = setTimeout(() => {
+        if (document.activeElement !== elChatinput) {
+          document.body.classList.toggle('showChat', false)
+        }
+      }, NotificationTime)
+    }
```
jdoleary, line 68: I'd rewrite this as just:
```
      if(chatTimeout){
          clearTimeout(chatTimeout);
      }
      chatTimeout = setTimeout(() => {
        if (document.activeElement !== elChatinput) {
          document.body.classList.toggle('showChat', false)
        }
      }, NotificationTime)
```
Rather than have the setTimeout block twice

src/views.ts
```diff
@@ -263,6 +267,11 @@ export function addOverworldEventListeners(overworld: Overworld) {
           }
         }
       },
+      {
+        target: elChatinput,
+        event: 'keypress',
```
jdoleary, line 272: If you wrap the input in a <form/> and give it `type="submit"` you can use the 'submit' event which will listen to Enter but not process other commands.  Then just remember to call `e.preventDefault()` in the event handler.  The benefit of this is that it wont listen to every keystroke.
Further reading: https://stackoverflow.com/a/27808062/4418836

src/graphics/ui/Chat.ts
```diff
@@ -0,0 +1,76 @@
...
+        chatname = globalThis.player.name;
+        chattercolor = globalThis.player?.color
+          ? hexToString(lightenColor(globalThis.player.color, 0.3))
+          : chattercolor;
+      }
+      message = elChatinput.value;
+      if (elChatinner) {
+        elChatinner.innerHTML += `<div class="message-speaker" style="color: ${chattercolor};">${chatname}:
+            <span class="text" style="color: white;">${message}</span></div> <br>`;
+      }
```
jdoleary, line 32: Rather than having this logic in both sendChatHandler and RecieveMessage just have it in Recieve message and then don't exclude the current player from processing their own message.  This way you only have to maintain one area where chat text gets added to the chatbox

src/graphics/ui/Chat.ts
jdoleary: I think it's because the event is still bubbling.  You may need to use e.stopImmediatePropagation() instead of disabling the input

jdoleary (commented):
Great work so far!

I'm still getting the initial "t" showing up in the chatbox whenever I open it. And I'm experiencing a little weirdness with the chatbox not going away.  It might be best to just have a "x" button to close the chat window if you don't want to see it instead of using a timeout.  Thoughts?

I left a few inline comments with some suggestions. Looking forward to hearing your thoughts!

src/graphics/ui/Chat.ts
jdoleary: This is the intent behind WebsocketPie, you can process messages you send and messages you receive in the same place :)

src/graphics/ui/Chat.ts
Couls: passed event to focusChat and used .preventDefault instead

src/graphics/ui/Chat.ts
Couls: Rewrote the section so all the logic is in receive message, and all we're sending is the message now

Couls:
chatbox is fixed, seems I lost some of my changes when undoing the lint and format they've been restored now, so clicking outside of chat will now cause it to disappear, and t won't appear chat when you first press it and I restored the style changes so they match the screenshot again

jdoleary:
It works and looks great!  I added a little input sanitization and have it scroll to the latest message when a new message is received.

Please just acknowledge in a comment that you agree to the [license](https://github.com/jdoleary/Spellmasons/blob/master/LICENSE.md) and I will be happy to merge this PR!

Couls:
Looks great! I agree to the license

jdoleary (approved):

Issue #86, opened 2023-12-09
Increase Stamina/HP/Range per point


SoulMuncher:
Putting stat points into mana is _infinitely_ more powerful than putting stat points elsewhere.

Mana is better at giving you endurance than health is.
20 mana or 32 health (less health than most enemies do damage per-attack)
- This is an extra blood curse, purify, heal, shield, and similar spell every turn.
- This is the difference between being able to resurrect the 1000 hp miniboss or tanking the entire horde of enemies yourself
- You can turn it into offense, to kill the enemy that would've otherwise hit you
- Decoy.

Mana is better at giving you mobility than stamina is. 
20 mana or 40 stamina (19% walking distance):
- This is an extra push, pull, repel, dash, swap to move somewhere
- This is an extra targeting spell to reach something
- This is an extra slash to slow, freeze, or kill the enemy chasing you

Mana is better at giving you range than cast range is:
20 mana or 12% cast range
- This is an extra cone, column, or even plus radius
- This is the push before you connect
- This is what lets you add target similar to your mid game spell to wipe out all the summoners

It is individually better at doing everything the other stats can do, while also having the versatility to switch between uses as the situation changes, AND it gives you more damage. It lets you kill Deathmason turn 1 and then target similar + debilitate + rend when he splits into 3.

On top of that, Mana is the only stat that doesn't have diminishing returns, and in fact has increasingly exponential returns. Every point of health, stamina, or range increases that stat by a smaller amount, and affects the game less than the last point did, while every 20 mana is another targeting spell, which often doubles your potential for damage, health, range, number of summons, or turn efficiency (letting you do that big combo all at once instead of over two turns)

-

While a huge part of this imbalance is due to the multiplicative nature of targeting spells, I do also think that the raw stat gains per point need significant adjustment.

In the long run, and alongside a targeting spell nerf, it might also be a good idea to increase the player's starting mana, so they aren't immediately forced to take it.

jdoleary:
Haha totally agree.  Let's definitely buff the stam/hp/range per point. 

As for the targeting nerf, I want to see how the changes I'm making to Target Similar and Target Kind affect that.  It'll be interesting to see if that makes "map-wipe" spells less common.  We can adjust gradually.

SoulMuncher:
@jdoleary 
I've been doing some testing, here are my initial thoughts:

[...]

PR #113, opened 2023-12-10
Increased non-mana Stat Gain

SoulMuncher:
#86

First thoughts posted in issue ^

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -30,7 +30,7 @@ export const UNIT_BASE_HEALTH = 40;
 export const UNIT_BASE_MANA = 60;
 export const UNIT_BASE_STAMINA = 300;
 // For game difficulty, I'm making the attack range less than the unit base stamina
-export const PLAYER_BASE_ATTACK_RANGE = 240; //previously 240 | UNIT_BASE_STAMINA * 0.8
+export const PLAYER_BASE_ATTACK_RANGE = UNIT_BASE_STAMINA * 0.8;
 // For game difficulty, player stamina less than the unit stamina so they can't run away without upgrading it
 export const PLAYER_BASE_STAMINA = 200; //previously 210 | UNIT_BASE_STAMINA * 0.7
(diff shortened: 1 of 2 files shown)


src/config.ts
```diff
@@ -30,9 +30,9 @@ export const UNIT_BASE_HEALTH = 40;
 export const UNIT_BASE_MANA = 60;
 export const UNIT_BASE_STAMINA = 300;
 // For game difficulty, I'm making the attack range less than the unit base stamina
-export const PLAYER_BASE_ATTACK_RANGE = UNIT_BASE_STAMINA * 0.8;
+export const PLAYER_BASE_ATTACK_RANGE = 200; //previously 240 | UNIT_BASE_STAMINA * 0.8
```
jdoleary, line 33: This will probably make the early game harder, but I think I like that.  It'll also make players much more inclined to choose range upgrades

jdoleary: I'll have to play with it a bit before approving, but intuitively, I like this change.

jdoleary: To add to my reasoning, I initially made player attack range 80% of unit base stamina so that it would be possible for golems to be out of your range when you end your turn and still hit you on their turn.  So you have to step into danger in order to target them.  This change increases that and I feel like new players find the game to be hard until they get the hang of it - so I'm more interested in making the game harder for established players than new players, but regardless, I will try this out and see what I think.  Just wanted to add some historical context

SoulMuncher: That makes sense! The very first level is definitely made more difficult with this change. Interesting to think about.

I agree the game should be more difficult for experienced players, and also agree that we don't want to scare new players off. I will have to keep that second part in mind when thinking about difficulty.

jdoleary: Agreed,  revert the PLAYER_BASE_ATTACK_RANGE change and keep the rest and I'll merge it 👍

jdoleary (approved):

Issue #109, opened 2023-12-09
globalThis.freeSpells should be player.freeSpells


jdoleary:
A player reported losing Summon spells.  I think this is because freeSpells is stored on the global and not on the player so it doesn't get serialized.

I need to reproduce this, but this may be why.

PR #114, opened 2023-12-10
Player Free Spells

SoulMuncher:
#109

### DIFF
--- a/src/cards/summon_generic.ts
+++ b/src/cards/summon_generic.ts
@@ -127,10 +127,10 @@ export default function makeSpellForUnitId(unitId: string, asMiniboss: boolean, 
     const expenseScaling = 2;
     const manaCost = override?.properties.manaCost ? (override.properties.manaCost * (asMiniboss ? 1.5 : 1)) : Math.max(1, Math.round(2 * Math.log2((sourceUnit.spawnParams?.budgetCost || 1)))) * 30 * (asMiniboss ? 2 : 1);
     const id = unitId + (asMiniboss ? ' Miniboss' : '');
-    if (!globalThis.freeSpells) {
-        globalThis.freeSpells = [];
+    if (globalThis.player && !globalThis.player?.freeSpells) {
+        globalThis.player.freeSpells = [];
     }
-    globalThis.freeSpells.push(id);
+    globalThis.player?.freeSpells.push(id);
     const unitSource = allUnits[unitId];
     let healthMax = unitSource?.unitProps.healthMax || config.UNIT_BASE_HEALTH;
     let manaMax = unitSource?.unitProps.manaMax || 0;
@@ -143,7 +143,7 @@ export default function makeSpellForUnitId(unitId: string, asMiniboss: boolean, 
 🗡️ ${unitSource.unitProps.damage} ${i18n(['damage'])}` : ''}${!!unitSource.unitProps.attackRange ? `
 🎯 ${unitSource.unitProps.attackRange} ${i18n(['attack range'])}` : ''}
 ❤️ ${healthMax} ${i18n(['health capacity'])}
-${manaMax ? `🔵 ${manaMax} + ${unitSource.unitProps.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}` : ''}`;
+    ${manaMax ? `🔵 ${manaMax} + ${unitSource.unitProps.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}` : ''}`;
 
     return {
         card: {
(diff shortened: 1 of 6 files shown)


src/cards/summon_generic.ts
```diff
@@ -143,7 +143,7 @@ export default function makeSpellForUnitId(unitId: string, asMiniboss: boolean,
 🗡️ ${unitSource.unitProps.damage} ${i18n(['damage'])}` : ''}${!!unitSource.unitProps.attackRange ? `
 🎯 ${unitSource.unitProps.attackRange} ${i18n(['attack range'])}` : ''}
 ❤️ ${healthMax} ${i18n(['health capacity'])}
-${manaMax ? `🔵 ${manaMax} + ${unitSource.unitProps.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}` : ''}`;
+    ${manaMax ? `🔵 ${manaMax} + ${unitSource.unitProps.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}` : ''}`;
```
jdoleary, line 146: Is this whitespace intentional? In a template literal string it will preserve the whitespace so this line will be out of alignment with the others.

jdoleary:
This looks good but I haven't tested it yet.  As part of our process can you please describe in the PR description what QA testing you did?
We'll also need to test saving and loading games with this PR.

src/cards/summon_generic.ts
SoulMuncher: This whitespace change is not intentional.

SoulMuncher:
@jdoleary 
> As part of our process can you please describe in the PR description what QA testing you did?

Yes, we can add that to the PR process. I'm not 100% sure it'd be worth the time I could spend fixing another bug, but I'm definitely willing to try it out for a bit. It might be helpful with larger game enhancements and more complex issues.

Just to give you a little more insight on my process: If I've posted a PR, I have already attempted fairly extensive QA and took the time to test every interaction that came to mind. It might be good to avoid the additional QA overlap. I may miss something from time to time, especially as I learn the nuances of the spellmasons-codebase, but I do try to test every change thoroughly and I feel I catch a pretty good number of issues pre-PR. Also, I think it would take a not-insignificant amount of time to explain every test I run, and a lot of tests probably take longer to explain than to do, haha.

SoulMuncher:
> We'll also need to test saving and loading games with this PR.
I did not test saving and loading. Good idea

jdoleary:
> @jdoleary
> 
> > As part of our process can you please describe in the PR description what QA testing you did?
> 
> Yes, we can add that to the PR process. I'm not 100% sure it'd be worth the time I could spend fixing another bug, but I'm definitely willing to try it out for a bit. It might be helpful with larger game enhancements and more complex issues.
> 
> Just to give you a little more insight on my process: If I've posted a PR, I have already attempted fairly extensive QA and took the time to test every interaction that came to mind. It might be good to avoid the additional QA overlap. I may miss something from time to time, especially as I learn the nuances of the spellmasons-codebase, but I do try to test every change thoroughly and I feel I catch a pretty good number of issues pre-PR. Also, I think it would take a not-insignificant amount of time to explain every test I run, and a lot of tests probably take longer to explain than to do, haha.

No need to explain QA in detail.  I definitely don't want to lose significant dev time in process.  I was more thinking a super quick:
"QA'd: save/load, multiplayer" or something like that.  Super simple, just to communicate higher-level of what was tested.

SoulMuncher:
> No need to explain QA in detail. I definitely don't want to lose significant dev time in process. I was more thinking a super quick: "QA'd: save/load, multiplayer" or something like that. Super simple, just to communicate higher-level of what was tested.

Gotcha, that's more than reasonable. Will add this to the process.

src/entity/Player.ts
```diff
@@ -122,10 +125,7 @@ export function changeMageType(type: MageType, player?: IPlayer, underworld?: Un
         {
           const upgrade = Upgrade.getUpgradeByTitle(arrowCardId);
           if (upgrade) {
-            underworld.chooseUpgrade(player, upgrade);
-            if (player == globalThis.player) {
-              globalThis.freeSpells.push(arrowCardId);
-            }
```
jdoleary, line 128: @SoulMuncher I almost missed this.  When this was changed from globalThis.freeSpells to player.freeSpells it still would've been a problem because _every_ client need to have the player.freeSpells.push happen so that all player objects are in parity.  Now getFreeUpgrade handles this.  The player object that is passed in gets the spell and all clients keep their player objects in parity.

jdoleary (approved):

Issue #91, opened 2023-12-09
Multi-Cast Contaminate


SoulMuncher:
Currently doesn't do anything.

What if it chained once for every additional cast? How would this work with the other upcoming changes?

jdoleary:
I think it should add stacks to the curses that spread.  So casting 3xContaminate on a unit with `poison 1` gives `poison 3` to the nearby enemies. 
My intention was the Plus Radius would be used to increase size, but maybe Plus Radius just needs to be buffed to be much more worth using.  It feels kinda weak now

SoulMuncher:
I'm working on the contaminate chaining and I ran into something:

4 units in a line: A - B - C - D

A - Slow 1
B - Slow 5
C - Slow 10
D - Slow 1

Cast 2x Contaminate on B

Which of these is the expected outcome?
- D gets slow 5 (because you're only spreading B's curses)
- D gets slow 10 (because it chains over C)
- B and D both get slow 10 (because C should also chain back to B)
- They all get slow 10 because *magic* (because everything is somewhere in the chain and they should all receive the highest-level curses, even though B > C > B > A would technically be 3 chains and thus require an extra contaminate cast)

jdoleary:
>  I like that some spells can kind've "replace" targeting spells with their built in mechanics. I like that I can currently play a witch with minimal targeting spell usage.
I like this too.  My intuition is that it spreads the target's curse only, so `D gets slow 5 (because you're only spreading B's curses)`

I would expect Contaminate chaining to make the curse hop like skipping a stone on a lake and spread it around based on the level of the target unit's curse

Issue #90, opened 2023-12-09
Multi-Cast Last Will doesn't do anything


SoulMuncher:
Currently doesn't do anything.

jd
> doesn't increase potion drops by design, it got too crazy with dropping huge numbers of mana potions - too easy to get huge mana; worth discussing

jdoleary:
Ref my comment on #93.  I'm open to ideas tho

SoulMuncher:
Last Will could use the new pickup strength feature

Issue #23, opened 2023-12-05
Revisit how sacrifice stacks


jdoleary:
> SoulMuncher:
> Multiple sacrifices increases your hp gained. I assume it's intentional but mentioning it since it's like... pretty darn powerful

I think sacrifice shouldn't take more health than the ally unit has, but maybe stacking it should have _some_ kind of interesting effect?

SoulMuncher:
After thinking through some ideas, my favorite one is simply making sacrifice steal a flat amount of hp, stacking per cast, and not giving you more life than each affected unit has. I think this is especially nice for higher difficulty/calamity balance purposes.

Some other ideas include:
- Steal half or a % of unit's hp, stacking per cast and never reaching 100%. May not be intuitive or that fun/worth stacking. Doesn't fix balance issues.
- Sacrifice could apply as a curse similar to suffocate, with the effect happening at the end of a timer. Kind've interesting imo. Like a ritual.
- Sacrifice could occur immediately and apply a curse similar to suffocate. One cast would instantly kill the unit, while multicasting would allow the unit to survive a couple more turns even after you gain health. You would not be able to re-sacrifice a unit while they're marked for death. May not be intuitive and may be imbalanced/problematic with purify.
- Sacrifice could be a poison-like curse that steals hp each turn. Stacking = greater amount or longer duration. I like this idea too.

- Health and "overshield" rebalance? Kind've a random thought: Should I be able to get 10000 health or mana if my max is only 100? Where is the line drawn, if ever.

jdoleary:
Good ideas - some seem like they'd make excellent spells of their own.
I want sacrifice to allow you to kill your ally - it's kinda fun and dark that way. But I agree that stacking needs to have a sensible effect.
I think I like the idea of it taking a flat amount, and if you exceed their health then they die (but it never takes more than they have).

> Health and "overshield" rebalance? Kind've a random thought: Should I be able to get 10000 health or mana if my max is only 100? Where is the line drawn, if ever.

I want it to remain absurd.  Obviously, we'll need to fix the exploit where you can take more health than exists, but if a player is smart enough to find a way to get 10000 health, I want them to feel rewarded for doing so.

PR #115, opened 2023-12-10
Spell Balancing and Fixes

### LINKED ISSUE #60: Dash + Burst
If you cast dash + burst on an enemy at medium distance, the prediction assumes you will complete the entire dash before casting burst, causing the unit to die, but in gameplay, you will often cast burst before dash finishes, causing you to deal less damage and leave the target alive.

### LINKED ISSUE #69: Buff Suffocate and Poison
Branching this off from #14 so it doesn't get forgotten.

### LINKED ISSUE #70: Increase arc angle when target cone is stacked
Maybe 50% more

### LINKED ISSUE #74: Allow Contaminate to increase stacks of already-held curses

### LINKED ISSUE #87: Repel Balance
Repel can easily substitute for most other movement spells with no additional cost.

Does it need a nerf?

While repel *is* very powerful and outshines the other movement spells, I think it's also very fun to use, so we should be careful in adjusting it.

### LINKED ISSUE #93: Manasteal Nerf
I hope I don't need to elaborate on this one

SoulMuncher:
I encourage you to take some time to relax and try out these changes! Witch + Suffocate is really fun.
Feel free to discuss/revert any of these changes, you can consider some of them "suggestions" more than anything.

 I think these will go well with the changes coming in PR #113 

Closes #70 
- Fixed an issue where a fully stacked cone would not be able to target anything
- Cone's angle now doubles each cast, up to 4 when it becomes a full circle
- Slightly adjusted mana cost and range of other targeting spells for balance consistency
- Fixed interaction between target column and plus radius that caused the width of the spell to grow too big too fast

Closes #87
- Rebalanced Repel and Vortex
- Repel and Vortex should now require Push and Pull respectively (I am leaning towards reverting the mana change and keeping this, personally)

Closes #69 
- Poison damage increased from 18 to 20
- Suffocate reworked: Suffocate will now build up exponentially over time until its buildup/damage is greater than the current health of the enemy, at which point the enemy dies.
- Suffocate buildup at x quantity (quantity increases by 1 per cast and 1 per turn):
10/14/**20**/28/**40**/
56/**80**/113/**160**/226/
**320**/452/**640**/905/**1280**
- Curses like poison and suffocate now "tick" at the end of the affected unit's turn instead of the beginning. Left some notes about that in discord.

Closes #74 
- Contaminate spreads to already infected units by overwriting lower-level curses.

#93 (changes reverted) 
- Manasteal now steals 30 mana instead of 20, but becomes less effective for each unit it hits. Hitting more units will always give you more mana overall, but you will steal less from each of them individually. It will steal up to 4 times its base amount of mana per quantity/cast, and is stolen evenly between enemies.
- Expense scaling increased from 1 to 2

Closes #60 
- Quick fix: Dash now has to await a promise, similar to other movement/force spells.

Other
- Slow amount increased from 0.8 * maxStamina to 0.75

Still missing
#90
#91
#23

TODO:
- Contaminate still needs a multicast function
- Prettier solution for cone angle/calc fix
- Make sure all relevant changes are reflected in spell info/descriptions


src/Underworld.ts
```diff
@@ -2210,7 +2210,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
       await Promise.all(unit.onTurnEndEvents.map(
         async (eventName) => {
           const fn = Events.onTurnEndSource[eventName];
-          return fn ? await fn(unit, this, false) : false;
+          return fn ? await fn(unit, false, this) : false;
```
jdoleary, line 2213: Thank you for fixing this inconsistency!

src/cards/contaminate.ts
```diff
@@ -48,14 +49,23 @@ async function spreadCurses(casterPlayer: IPlayer | undefined, unit: IUnit, unde
...
     // Only curses are contagious
     // Do not spread contaminate itself
-    .filter(([cardId, modValue]) => modValue.isCurse && cardId !== contaminate_id)
+    .filter(([cardId, modValue]) => modValue.isCurse && cardId !== contaminate_id && cardId !== impendingDoomCardId)
     .map(([id, mod]) => ({ card: allCards[id], quantity: mod.quantity }))
     .filter(x => x.card !== undefined) as { card: ICard, quantity: number }[];
 
   for (let { card, quantity } of curseCardsData) {
     const promises = [];
     // Filter out units that already have this curse
```
jdoleary, line 58: @SoulMuncher : Please update this comment to align with modified logic below

src/cards/dash.ts
```diff
@@ -18,11 +18,13 @@ const spell: Spell = {
     thumbnail: 'spellIconDash.png',
     description: 'spell_dash',
     effect: async (state, card, quantity, underworld, prediction) => {
+      let promises = [];
       const targets = getCurrentTargets(state);
       playDefaultSpellSFX(card, prediction);
       if (targets[0]) {
-        pull(state.casterUnit, targets[0], quantity, underworld, prediction);
+        promises.push(pull(state.casterUnit, targets[0], quantity, underworld, prediction));
```
jdoleary, line 25: This could be written more simply as just `await pull(...`

No need for Promise.all(array) if there's only one.

Note: Dash used to dash to each target, in order, but this caused a significant multiplayer desync.  Because of the way forceMoves are handled on the headless server and in prediction, each spell can only cause a single `await` to occur (or else it will desync)
So a single `await Promise.all` works and a single `await promise` works but there can't be multiple `awaits` nor an `await` inside of a for loop.
When I changed Dash to fix this, I accidentally removed the await entirely which is what created the bug that you're fixing here.
But this history is important to know for future forceMove changes.  

If you're curious to learn more, see `triggerGameLoopHeadless` and `fullySimulateForceMovePredictions`, they're only invoked once after each spell effect.

src/cards/freeze.ts
```diff
@@ -104,6 +104,7 @@ const spell: Spell = {
           }
         }
       }
+      return false;
```
jdoleary, line 107: I think returning a boolean for onTurnEnd shouldn't be necessary, the boolean for onTurnStart is only needed for skipping the unit's turn.  For example, freeze returns true so that that unit wont take a turn while they're frozen

src/cards/impending_doom.ts
```diff
@@ -0,0 +1,106 @@
...
+  }
+}
+export function updateTooltip(unit: Unit.IUnit) {
+  if (unit.modifiers[impendingDoomCardId]) {
+    // Set tooltip:
+    unit.modifiers[impendingDoomCardId].tooltip = `Impending doom ${unit.modifiers[impendingDoomCardId].quantity}...`
+  }
+}
+
+//Dev spell for toxic res: Unit dies in specified number of turns
```
jdoleary, line 41: Is this just a copy of suffocate for toxic res? If so, it doesn't need to be a whole spell, it can just be a modifier.  See src/modifierSummoningSickness and src/cards/index.ts line 272 (registerSummoningSickness()) and priest.ts line 41 for how to use/create modifiers that aren't attached to a card

src/cards/index.ts
```diff
@@ -104,6 +105,7 @@ import { registerUrnpoisonExplode } from '../entity/units/urn_poison';
 import { registerUrnexplosiveExplode } from '../entity/units/urn_explosive';
 import { calculateGameDifficulty } from '../Difficulty';
 import registerCorpseDecay from '../modifierCorpseDecay';
+import { over } from 'lodash';
```
jdoleary, line 108: Typo import?

src/cards/mana_steal.ts
```diff
@@ -21,18 +22,18 @@ const spell: Spell = {
...
+    description: ['spell_mana_steal', base_mana_stolen.toString()],
     effect: async (state, card, quantity, underworld, prediction) => {
       // .filter: only target living units
       const targets = state.targetedUnits.filter(u => u.alive && u.mana > 0);
       const caster = state.casterUnit;
       let promises = [];
       let totalManaStolen = 0;
       for (let unit of targets) {
-        const unitManaStolen = Math.min(unit.mana, mana_stolen * quantity);
+        const unitManaStolen = Math.min(unit.mana, Math.floor(base_mana_stolen * quantity / (((targets.length - 1) / max_mana_stolen_mult) + 1)));
```
jdoleary, line 36: I'm hesitant about this one for reasons I mentioned earlier.  I'll test it out in game and see how it feels.

jdoleary: Similar to the balancing I recently did to Target Similar, it needs to allow "super awesome ridiculous plays" but it shouldn't be too easy to do so.  I'm all for sensible balancing of the spells that are too easy to use powerfully but we must maintain that Spellmasons is a game where you can take magic to absurd levels - that's core to it's identity.
Let's talk more about it next time we chat

src/cards/repel.ts
```diff
@@ -1,18 +1,19 @@
...
-import { forcePush } from './push';
+import { forcePush, pushId } from './push';
 import { CardRarity, probabilityMap } from '../types/commonTypes';
 
 export const id = 'repel';
 export const velocityStartMagnitude = 10;
 const spell: Spell = {
   card: {
     id,
+    requires: [pushId],
```
jdoleary, line 12: 100%! Love this change

src/cards/repel.ts
```diff
@@ -1,18 +1,19 @@
...
     category: CardCategory.Movement,
     supportQuantity: true,
     sfx: 'push',
-    manaCost: 10,
+    manaCost: 20,
```
jdoleary, line 16: You mentioned considering reverting the mana change.  Maybe it should be 15?  I like keeping movement spells cheap, but if it's 10 then there's literally no reason to ever use push again.

src/cards/suffocate.ts
```diff
@@ -24,13 +25,19 @@ function add(unit: Unit.IUnit, underworld: Underworld, prediction: boolean, quan
...
   });
-  const lastTurnsLeftToLive = modifier.turnsLeftToLive;
-  modifier.turnsLeftToLive = 1 + Math.ceil(unit.health / 20 / modifier.quantity)
-  // Ensure that casting Suffocate doesn't ever increase the turns left
-  if (lastTurnsLeftToLive) {
-    modifier.turnsLeftToLive = Math.min(modifier.turnsLeftToLive, lastTurnsLeftToLive);
+
+  modifier.buildup = Math.floor(10 * Math.pow(2, (modifier.quantity - 1) / 2));
+
+  if (modifier.buildup >= unit.health) {
```
jdoleary, line 31: This is very interesting because it means suffocate is more powerful if you hurt them and less powerful if you heal them.  I think I like it.
TODO: I'll need to make localization changes to any of these spells when their functionality changes.

src/cards/target_circle.ts
```diff
@@ -11,13 +11,13 @@ import { CardRarity, probabilityMap } from '../types/commonTypes';
 import { HasSpace } from '../entity/Type';
 
 const id = 'Target Circle';
-const baseRadius = 140;
+const baseRadius = 100;
```
jdoleary, line 14: Reasoning for the nerf?

src/cards/target_column.ts
```diff
@@ -31,7 +31,7 @@ const spell: Spell = {
     allowNonUnitTarget: true,
     effect: async (state, card, quantity, underworld, prediction, outOfRange) => {
       const depth = (range + state.aggregator.radius) * (0.5 + 0.5 * quantity);
-      const width = (baseWidth + state.aggregator.radius) * (0.5 + 0.5 * quantity);
+      const width = baseWidth * Math.pow(2, Math.min(quantity, 4)) / 2;
```
jdoleary, line 34: A comment would be helpful here so I don't have to mentally process the math to understand how or why it scales this way

src/cards/target_cone.ts
```diff
@@ -29,8 +29,8 @@ const spell: Spell = {
     description: 'spell_target_cone',
     allowNonUnitTarget: true,
     effect: async (state, card, quantity, underworld, prediction, outOfRange) => {
-      const adjustedRange = range + state.aggregator.radius;
-      const adjustedAngle = coneAngle * quantity;
+      const adjustedRange = (range + state.aggregator.radius) * (0.75 + 0.25 * quantity);
+      const adjustedAngle = coneAngle * Math.pow(2, Math.min(quantity, 4)) / 2;
```
jdoleary, line 33: Same here, a comment would be helpful to understand the math at a glance

src/cards/target_cone.ts
```diff
@@ -71,8 +71,11 @@ function withinCone(origin: Vec2, coneStartPoint: Vec2, radius: number, startAng
   // and within angle:
   const targetAngle = getAngleBetweenVec2s(coneStartPoint, target);
   const distanceToConeStart = distance(target, coneStartPoint);
+
+  //TODO - Investigate isAngleBetweenAngles
+  //temp fix for cone inversion: if angle is whole circle, just check distance.
```
jdoleary, line 76: Agreed, just switch to distance if it's a whole circle

src/cards/target_similar.ts
```diff
@@ -19,7 +19,7 @@ const spell: Spell = {
     id: targetSimilarId,
     category: CardCategory.Targeting,
     supportQuantity: true,
-    manaCost: 20,
+    manaCost: 25,
```
jdoleary, line 22: 👍

src/cards/target_similar_2.ts
```diff
@@ -10,7 +10,7 @@ const spell: Spell = {
     replaces: [targetSimilarId],
     category: CardCategory.Targeting,
     supportQuantity: true,
-    manaCost: 20,
+    manaCost: 25,
```
jdoleary, line 13: 👍

src/cards/vortex.ts
```diff
@@ -1,17 +1,18 @@
...
 import { playDefaultSpellSFX } from './cardUtils';
-import { pull } from './pull';
+import { pull, pullId } from './pull';
 import { CardRarity, probabilityMap } from '../types/commonTypes';
 
 export const id = 'vortex';
 const spell: Spell = {
   card: {
     id,
+    requires: [pullId],
```
jdoleary, line 11: 👍

src/cards/vortex.ts
```diff
@@ -1,17 +1,18 @@
...
     category: CardCategory.Movement,
     supportQuantity: true,
     sfx: 'pull',
-    manaCost: 10,
+    manaCost: 20,
```
jdoleary, line 15: Same as my comment for repel, I think 15 would keep it independent from pull but have it remain cheap so it can be used strategically as a movement spell.

jdoleary (changes requested):
Great work!  Super excited about these changes.  Left a few comments and change requests

src/cards/dash.ts
SoulMuncher: That's actually really good information. Thanks!

src/cards/index.ts
SoulMuncher: Never seen that before. Lol.

src/cards/suffocate.ts
SoulMuncher: We could always change it to use max hp too, if you don't want it to work like a 2nd poison. I think it's nice either way.

src/cards/target_circle.ts
SoulMuncher: The radius decrease was done in conjunction with reducing the mana cost. I actually thought this was a pretty nice buff to circle overall, so I'm curious if you'll feel the same after trying it out.

The main reason for these changes is that they're an extension of the cone change. Cone had become better at being a circle than circle was, due to circle's incredibly high base mana cost. So I decreased it, and adjusted the radius accordingly. This means it might be weaker at one single cast, but as you add more you should end up with a bigger circle for the same amount of mana. (And is arguably better at single casting, since it's cheaper and the entire base radius may not be needed)

This brings it more in line with the other targeting spells and should make it a little more viable/versatile/interesting. The goal is to make multi-casting more worthwhile, allow players to use smaller or bigger targeting options as they need to, and make sure circle has a use before super-endgame.

src/cards/mana_steal.ts
SoulMuncher: Understandable. I'm honestly not 100% on board with this change either. Balance wise, it feels awesome, but the formula is a little arbitrary and unintuitive, I think. It's not something you can really understand or calculate quickly without a graph in front of you.

Ideally, it'd be as easy to understand as "steal 20 mana from every target", while also not being so strong it makes all other game strategies redundant. My main problem with mana steal isn't the absurdity of it, but the fact that it's so easy to execute. Imo it feels more like a cop out or cheat than a ridiculously awesome and creative play in its current state, and that's primarily what I want to fix.

jdoleary: Yeah, I feel that.
I tried this one out and it feels really nerfed now.  Lets revert this for now and put a pin in it to visit later.

src/cards/resurrect_toxic.ts
```diff
@@ -42,13 +43,8 @@ const spell: Spell = {
...
 
-          // TOXIC add suffocate 
-          Unit.addModifier(unit, suffocateCardId, underworld, prediction, 1);
-          const modifier = unit.modifiers[suffocateCardId];
-          if (modifier) {
-            modifier.turnsLeftToLive = turnsLeftToLive;
-            updateTooltip(unit);
-          }
+          // TOXIC add impending doom 
+          Unit.addModifier(unit, impendingDoomId, underworld, prediction, turnsLeftToLive * quantity);
```
jdoleary, line 47: I like this quantity change! 👍

src/cards/resurrect_toxic.ts
```diff
@@ -42,13 +43,8 @@ const spell: Spell = {
...
           Unit.resurrect(unit);
 
-          // TOXIC add suffocate 
-          Unit.addModifier(unit, suffocateCardId, underworld, prediction, 1);
-          const modifier = unit.modifiers[suffocateCardId];
-          if (modifier) {
-            modifier.turnsLeftToLive = turnsLeftToLive;
-            updateTooltip(unit);
-          }
+          // TOXIC add impending doom
```
jdoleary, line 46: A comment here would best describe why.
"Impending doom is added to kill units after a specified number of turns.  This is the distinguishing characteristic of Toxic Resurrect, it is weaker than resurrect because it doesn't last."

src/cards/target_circle.ts
jdoleary: Excellent reasoning!

src/modifierSummoningSickness.ts
```diff
@@ -42,7 +42,7 @@ export default function registerSummoningSickness() {
             // Skip turn
             return true;
         },
-        onTurnEnd: async (unit: Unit.IUnit, underworld: Underworld) => {
+        onTurnEnd: async (unit: Unit.IUnit, prediction: Boolean, underworld: Underworld) => {
```
jdoleary, line 45: Boolean should be lowercase.  In js Boolean and boolean are actually different and so is String and string (always prefer the lower case).  It's weird I know... lol

jdoleary (approved):

PR #131, opened 2023-12-11
Spell Balancing 2

### LINKED ISSUE #23: Revisit how sacrifice stacks

### LINKED ISSUE #77: Can send mana to self (indefinitely)
Send mana should not be able to target the caster.

### LINKED ISSUE #86: Increase Stamina/HP/Range per point

### LINKED ISSUE #91: Multi-Cast Contaminate

### LINKED ISSUE #118: Impending Doom + Contaminate
I haven't tested but I just realized that when I changed impending doom from a "spell" to a modifier, it lost its “curse” classification, meaning it won't be removed by purify anymore.

What do we want to do about this?
- Make it a curse for the purposes of purify, contaminate, etc.
- Leave it the way it is and consider it a nerf to Toxic Res?
- Something else?

### LINKED ISSUE #122: Timemason + Send Mana exploit
> Coul — Yesterday at 10:24 PM
> send mana in multi can be spammed out in a few seconds at the start of every round to give the multiplayer ally timemason mana pool without worrying about the time mechanic

### LINKED ISSUE #129: Send Mana should have some expense scaling
so some mana is lost over time.
Ref: https://discord.com/channels/1032294536640200766/1032294537235812404/1183610146186874990

SoulMuncher:
Closes #91 
- Contaminate now "chains" with multicast

Closes #23 
- Sacrifice now deals 40 damage per quantity, stealing an equal amount of health but not exceeding the targeted unit's current health.

Closes #118 
- Impending Doom can now be spread via contaminate
- This, like other modifiers, overwrites similar modifiers with less quantity. This means that impending doom is actually the only modifier that overwrites stronger versions of the curse with weaker ones, as more quantity means more turns until death in this case.

Closes #86 
- Player starts with 60 hp and 200 range

Closes #77 
Closes #122
Closes #129 
- Send mana now has default expense scaling

QA'd singleplayer spell interactions

### DIFF
--- a/src/cards/contaminate.ts
+++ b/src/cards/contaminate.ts
@@ -18,6 +18,7 @@ const spell: Spell = {
   card: {
     id: contaminate_id,
     category: CardCategory.Curses,
+    supportQuantity: true,
     manaCost: 50,
     healthCost: 0,
     expenseScaling: 1,
@@ -27,7 +28,7 @@ const spell: Spell = {
     effect: async (state, card, quantity, underworld, prediction) => {
       // .filter: only target living units
       for (let unit of state.targetedUnits.filter(u => u.alive)) {
-        await spreadCurses(state.casterPlayer, unit, underworld, state.aggregator.radius, prediction);
+        await contaminate(state.casterPlayer, unit, underworld, state.aggregator.radius, prediction, quantity);
       }
       return state;
     },
@@ -35,16 +36,47 @@ const spell: Spell = {
 };
 export default spell;
 
-async function spreadCurses(casterPlayer: IPlayer | undefined, unit: IUnit, underworld: Underworld, extraRadius: number, prediction: boolean) {
+// separate function to handle synchronous recursion and animation - avoids long wait times
+async function contaminate(casterPlayer: IPlayer | undefined, unit: IUnit, underworld: Underworld, extraRadius: number, prediction: boolean, quantity: number) {
   const range = (COLLISION_MESH_RADIUS * 4 + extraRadius) * (casterPlayer?.mageType == 'Witch' ? 1.5 : 1);
+
+  // the units to spread contaminate from, initally just unit, then only the latest additions to the chain
+  let nextUnits: IUnit[] = [];
+  let ignore: IUnit[] = []
+  nextUnits.push(unit);
+  ignore.push(unit);
+
+  let recursion = 1;
+  while (recursion <= quantity) {
+    const promises = [];
+    for (let nextUnit of nextUnits) {
+      promises.push(spreadCurses(nextUnit, range, underworld, prediction, ignore));
+    }
+    nextUnits = [];
+    let affectedUnitsArrays = await Promise.all(promises);
+
+    for (let affectedUnits of affectedUnitsArrays) {
+      //add all affected units, but make sure there are no duplicates
+      nextUnits = nextUnits.concat(affectedUnits.filter(u => !nextUnits.includes(u)));
+      ignore = ignore.concat(affectedUnits.filter(u => !ignore.includes(u)));
+    }
+    recursion += 1;
+  }
+}
+
+async function spreadCurses(unit: IUnit, range: number, underworld: Underworld, prediction: boolean, ignore: IUnit[]): Promise<IUnit[]> {
   drawUICircle(unit, range, colors.targetingSpellGreen, 'Contagion Radius');
-  const nearByUnits = underworld.getUnitsWithinDistanceOfTarget(unit, range, prediction)
+
+  const nearbyUnits = underworld.getUnitsWithinDistanceOfTarget(unit, range, prediction)
     // Filter out undefineds
     .filter(x => x !== undefined)
     // Do not spread to dead units
     .filter(x => x?.alive)
     // Filter out self
-    .filter(x => x !== unit) as IUnit[];
+    .filter(x => x != unit)
+    // Filter out other ignored units
+    .filter(x => !ignore.includes(x)) as IUnit[];
+
   const curseCardsData: { card: ICard, quantity: number }[] = Object.entries(unit.modifiers)
     // Only curses are contagious
     // Do not spread contaminate itself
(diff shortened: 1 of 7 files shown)


jdoleary:
All spells must be QA'd at least once on multiplayer (historically, some spells would cause huge server problems - crashes, infinite loops, etc).  Server logic is slightly different since it doesn't have a visual layer so there can be unexpected impacts.
Use `npm run headless` in a terminal to run the server.  If you haven't done this yet I can walk you through it tomorrow.

I don't have time tonight for a full review, but I look forward to checking this one out tomorrow!

src/cards/contaminate.ts
```diff
@@ -27,65 +30,103 @@ const spell: Spell = {
     effect: async (state, card, quantity, underworld, prediction) => {
       // .filter: only target living units
       for (let unit of state.targetedUnits.filter(u => u.alive)) {
-        await spreadCurses(state.casterPlayer, unit, underworld, state.aggregator.radius, prediction);
+        contaminate(state.casterPlayer, unit, underworld, state.aggregator.radius, prediction, quantity);
```
jdoleary, line 33: I believe this still must be `await`ed.  Or else if this is cast in sequence it will finish asynchronously after the next spell in the spell order instead of before it.

src/cards/send_mana.ts
```diff
@@ -23,7 +23,7 @@ const spell: Spell = {
     supportQuantity: true,
     manaCost: amount,
     healthCost: 0,
-    expenseScaling: 0,
+    expenseScaling: 1,
```
jdoleary, line 26: This might need more scaling than 1, we'll have to play with the balance of it.

src/cards/suffocate.ts
```diff
@@ -26,6 +26,8 @@ function add(unit: Unit.IUnit, underworld: Underworld, prediction: boolean, quan
     }
   });
 
+  // One suffocate stack is added per cast and per turn passed
+  // Buildup doubles every 2 stacks until > hp, then unit dies
```
jdoleary, line 30: Very helpful context!

jdoleary (changes requested):
Looks mostly good.  A few important comments in contaminate.ts and needs multiplayer testing to ensure there are no unexpected issues when run on the headless server.

src/cards/contaminate.ts
SoulMuncher: Oh true! I wanted all contaminates to go at the same time, because otherwise it takes suuuper long. I need to turn this into a promise[] I think, and await full completion. Thanks for pointing that out!

SoulMuncher:
Last commit closes these issues:
Closes #133 
Closes #134

SoulMuncher:
Last commit
Closes #136

src/cards/split.ts
```diff
@@ -78,21 +79,23 @@ function add(unit: Unit.IUnit, underworld: Underworld, prediction: boolean, quan
...
     }
   });
-  if (modifier.quantity && modifier.quantity >= splitLimit) {
-    return;
-  }
-  if (unit.image) {
-    unit.image.sprite.scale.x *= scaleMultiplier;
-    unit.image.sprite.scale.y *= scaleMultiplier;
+  for (let i = 0; i < quantity; i++) {
+    if (modifier.quantity && modifier.quantity > splitLimit) {
```
jdoleary, line 83: What if the quantity should bring you _to_ the split limit?  If a unit has no split, and you cast Split 5 it should bring them to split 2, not return and do nothing

jdoleary: Nevermind, I just tested and it works fine

jdoleary:
contaminate
This is super satisfying!
Some minor nit-picks:
- Let's make non targeting circles a different color than green so there isn't any confusion if contaminate and target circle, for example are used together.  Flame Strike (in the mod) uses red cause it deals damage, I think non targeting spread effects could be white.
- Let's make sure the circle does not appear while the spell is animating, only in prediction.

Other than that it looks great and was super fun to watch the green blobs chain one by one!

SoulMuncher:
@jdoleary
I can't see the old comment on the split return issue, but you were 100% correct that the code would return incorrectly given certain inputs. It doesn't actually happen in game because the spell applies it one by one, and contaminate only adds the difference between the source's modifier quantity and the target's modifier quantity, thus never going over the cap to return anyway. So... while it's not technically possible to do it in the current game game, if there WAS a source of applying >3 split or overcapping it, the add function would without at least splitting up to the cap. _I think this has been fixed_, but I'd have to create a method of overcapping split to actually QA it properly.

SoulMuncher:
> super fun to watch the green blobs chain one by one!

You should've seen how it looked when it was recursive and could chain back to previous enemies. It looked like they were playing catch! I watched it bounce back and forth along a huge line of enemies for a couple minutes...

Better visuals added. It's using the default red aoe color now. Maybe I should've added a darker green color, though.

SoulMuncher:
And fun note about contaminate:

It'll spread curses in the order that they initially appear on the first unit. This can make a big difference with bloat/suffocate. There are some interesting ways to make it spread further by adding just the right amount of these two curses, and using the push from bloat to "move" units that have been targeted by one of the contaminate chains. I also made sure the prediction accounts for this, so it's pretty neat!

PR #139, opened 2023-12-12
Keep on Death property

### LINKED ISSUE #135: Infinite Split + Resurrect
Split is removed but not reverted when a unit dies, which means you can resurrect them and split them again indefinitely, making them smaller than a pixel with all stats set to 1.

### LINKED ISSUE #137: Keep modifiers on death
There is no way to force a unit to keep a modifier through death. Maybe it would be worth adding a boolean in the Modifier interface or changing this default behavior.

I'm looking at Split (#135 ), Contaminate, Vampires (units with innate/persistent status effects), future possibilities like a cheat death blessing that triggers on the dead unit's next turn start, corpse explosions, etc.

SoulMuncher:
Closes #137 
- Modifiers now have an optional keep on death property (currently used for split and innate vampire blood curse)

Closes #135 
- Split isn't removed on death

QA - Singleplayer

I needed to merge #131 from master to make sure #135 was fixed, since I made this branch earlier. I don't merge often, so you may want to double check that I didn't break anything, lol.

### DIFF
--- a/src/entity/Unit.ts
+++ b/src/entity/Unit.ts
@@ -763,7 +763,9 @@ export function die(unit: IUnit, underworld: Underworld, prediction: boolean) {
   // Note: This must come AFTER onDeathEvents or else it will remove the modifier
   // that added the onDeathEvent and the onDeathEvent won't trigger
   for (let [modifier, _modifierProperties] of Object.entries(unit.modifiers)) {
-    removeModifier(unit, modifier, underworld);
+    if (!_modifierProperties.keepOnDeath) {
+      removeModifier(unit, modifier, underworld);
+    }
   }
 
   if (globalThis.player && globalThis.player.unit == unit) {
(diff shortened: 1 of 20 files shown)


src/entity/Unit.ts
```diff
@@ -763,7 +763,9 @@ export function die(unit: IUnit, underworld: Underworld, prediction: boolean) {
   // Note: This must come AFTER onDeathEvents or else it will remove the modifier
   // that added the onDeathEvent and the onDeathEvent won't trigger
   for (let [modifier, _modifierProperties] of Object.entries(unit.modifiers)) {
```
jdoleary, line 765: A leading "_" denotes that the variable is unused.  What's happening here is called "Destructuring" and it works with both objects and arrays:
```
let a = [0,1,2,3];
let [zero, one] = a;
```
```
let a = {health:10,mana:2};
let {health, mana} = a;
```
In this example, I was just extracting the modifier key and the properties were unused.   Since the properties are now used, it is conventional to remove the leading "_"

jdoleary (changes requested):
One small change requested

jdoleary:
#135 is indeed fixed, I just tested it.
Also git is pretty good with merges, if there are no explicit merge conflicts then nothing will be broken due to a merge.

SoulMuncher:
> Also git is pretty good with merges, if there are no explicit merge conflicts then nothing will be broken due to a merge.
There was a merge conflict in split add(), but I think I sorted it out 😄 

Naming convention fixed

jdoleary (approved):

Issue #98, opened 2023-12-09
Bloat radius on select


SoulMuncher:
You should be able to see the explosion radius of a unit affected by bloat when you select (not target) them.

Edit: Meaning a circle around them in the game world, not the numerical radius stat.

jdoleary:
What about other curses?  If a player has all the curses, which should show on select and which shouldn't?  This could be useful, but might not be necessary until they're predicted to die as it does now.  I'm on the fence on this one.

SoulMuncher:
@jdoleary 
> What about other curses? ... I'm on the fence on this one.

Updated original post for clarity. Most curses and status effects don't have anything to show outside of the stat panel: Slow, Freeze, Suffocate, Poison, Blood Curse, Shield, Conserve, and Last Will. They don't and still wouldn't do anything on screen because they have no reason to. Bloat does, and I think that it's explosion radius should be visible to the player as most other ones are. (Exploding arrow, Urns, etc.)

In the current version of the game, I don't there's a way to see how big the explosion radius of bloat is. It feels pretty bad, especially when you add multiplayer desync and somewhat broken predictions. There isn't a reliable way to know if you're going to damage the things you're trying to damage, and not damage the things you don't want to.

jdoleary:
Okay I'm open to this. 
It _should_ always show the bloat circle when a unit is predicted to take fatal damage by a spell... now that I think about it, it _doesn't_ show the circle if it will take fatal damage indirectly (via poison for example).  So yes, let's try showing the radius on select 👍

Issue #96, opened 2023-12-09
Background for hp/mp bars


SoulMuncher:
A small transparent background for the hp/hp bars above units' heads should make them easier to see.

jdoleary:
Some RTSs have a feature where only damaged units have their healthbar showing.  Might be worth considering.

PR #146, opened 2023-12-13
Doodads Refactor

### LINKED ISSUE #71: Urns lose "targeting tint" when they die in prediction mode

### LINKED ISSUE #97: Urn explosion radius on select
You should be able to see the explosion radius of an urn when you select it.

SoulMuncher:
Closes #97 
- Urn shows explosion radius when selected
- Going to revisit this code in broader context, and look for a solution to #98 

Closes #71 
- Urn corpses are now correctly removed, fixing tint and blood trail issues

QA Single/Multi/Loading


SoulMuncher:
I wonder if there's a better way to do Unit.SubType

Unit.SubType is essentially used to define a bunch of specific behaviors for similar types of units. Doodads might have a lot of different behaviors.

Maybe instead of specifying broader types, we specify the behaviors themselves, such as unit targeting style, whether or not it leaves a corpse, what targeting category it's in and so on.

jdoleary:
> I wonder if there's a better way to do Unit.SubType
> 
> Unit.SubType is essentially used to define a bunch of specific behaviors for similar types of units. Different doodads might have a lot of different behaviors.
> 
> Maybe instead of specifying broader types, we specify the behaviors themselves, such as unit targeting style, whether or not it leaves a corpse, what targeting category it's in and so on.

That is a very interesting idea.  I think i'd have to see the overlaps laid out to feel like a refactor like that would be worth it tho.  For now it's basically attack style with a few exceptions.

SoulMuncher:
@jdoleary 
I made a bunch of other changes stemming off of this branch, and now that I've made them I realize they probably should've been a separate branch/pr since it's a pretty big one and addresses a lot of unrelated issues. I'll wait until this is merged and open up a new PR branched off of the new "master". It's a big one 👀

SoulMuncher:
These last few commits greatly improve many game visuals.

- Added a new unit event: onDrawSelected. This lets you define specific drawing behaviors for different units, such as drawing a line for LOS units or an effect radius for doodads. What's extra cool is this also lets you attach modifiers with their own drawing properties (such as a bloat explosion radius) to appear when the unit is selected.
- Units now show a bloat explosion radius when selected
- Fixed an issue that cause most floating text, unit markers, and other "world UI" to leave the screen border at various zoom levels
- Fixed an issue that caused some prediction visuals and their label text to flicker when moving the mouse
- Clarity of floating text has been increased significantly (still room to change the colors and dropdown shadow if requested)
- Centralized a lot of draw functions, improving visual consistency and code maintainability

Closes #98 

IMPORTANT: The new unit event causes old saves to not load properly. This probably needs an exception handler or something. Let me know what I need to do for that, if anything 👍 

QA'd Single/Multi/Loading

SoulMuncher:
Closes #96
- Unit Hp/Mp bars now show "healing"
- Also have transparent backgrounds

src/entity/Unit.ts
```diff
@@ -1444,6 +1450,91 @@ export function findLOSLocation(unit: IUnit, target: Vec2, underworld: Underworl
...
+
+// handles drawing attack range, bloat radius, and similar graphics each frame while a unit is selected
+export async function drawSelectedGraphics(unit: IUnit, prediction: boolean = false, underworld: Underworld): Promise<void> {
+
+  if (!prediction) {
+    for (let drawEvent of unit.onDrawSelectedEvents) {
+      if (drawEvent) {
+        const fn = Events.onDrawSelectedSource[drawEvent];
+        if (fn) {
+          await fn(unit, prediction, underworld);
```
jdoleary, line 1462: Does this event function need to be async?

src/graphics/PlanningView.ts
```diff
@@ -43,31 +43,33 @@ export function initPlanningView() {
...
+      labelText.style.fontSize = 100;
       labelText.anchor.x = 0.5;
-      labelText.anchor.y = 0;
+      labelText.anchor.y = 0.5;
+      labelText.scale.x = 0.3;
+      labelText.scale.y = 0.3;
       containerUI.addChild(labelText);
     }
     if (mouseLabelText) {
+      mouseLabelText.style.fontSize = 100;
```
jdoleary, line 54: The floating text looks great! All it took was increasing the font size and shrinking the scale??

src/entity/Unit.ts
```diff
@@ -1444,6 +1450,91 @@ export function findLOSLocation(unit: IUnit, target: Vec2, underworld: Underworl
 
 }
 
+
+// handles drawing attack range, bloat radius, and similar graphics each frame while a unit is selected
+export async function drawSelectedGraphics(unit: IUnit, prediction: boolean = false, underworld: Underworld): Promise<void> {
+
+  if (!prediction) {
```
jdoleary, line 1457: Let's have this exclude globalThis.headless too.  It shouldn't cause any errors but there's no reason for the headless server to have to take any of the branches in this function since it's visual only.

src/graphics/PlanningView.ts
```diff
@@ -483,88 +335,168 @@ export function drawHealthBarAboveHead(unitIndex: number, underworld: Underworld
...
-            const imagePath = globalThis.player && u.faction === globalThis.player.unit.faction ? 'badgeDeathAlly.png' : 'badgeDeath.png';
-            ImmediateMode.draw(imagePath, skullPosition, (1 / zoom) + (Math.sin(Date.now() / 500) + 1) / 3);
+      // Draw base health bar
+      if (globalThis.unitOverlayGraphics) {
+        const healthBarColor = u.faction == Faction.ALLY ? colors.healthAllyGreen : colors.healthRed;
+        const healthBarHurtColor = u.faction == Faction.ALLY ? colors.healthAllyDarkGreen : colors.healthDarkRed;
+        const healthBarHealColor = u.faction == Faction.ALLY ? colors.healthAllyBrightGreen : colors.healthBrightRed;
+
+        const healthBarMax = u.healthMax || 1;
+        //background
```
jdoleary, line 347: The healthbar background seems to apply to the player but not to golems

jdoleary: Edit: I was wrong, it does show on NPCs but for the player the background color is the samee as the "Prediction Loss Health" color whereas for the NPCs it's not.  I think it'd be best if there were a different color for each.  For allies, real health is the bright green, prediction loss health is the dark green and then missing health could be the very dark grey (non colored). And for enemies real health is red, prediction loss health is dark red and missing health (background) is dark grey?

src/graphics/PlanningView.ts
```diff
@@ -483,88 +335,168 @@ export function drawHealthBarAboveHead(unitIndex: number, underworld: Underworld
...
+  const zoom = getCamera().zoom;
+  // 1/zoom keeps the attention marker the same size regardless of the level of zoom
+  // Math.sin makes the attention marker swell and shink so it grabs the player's attention
+  // + 1 makes it go from 0 to 2 instead of -1 to 1
+  // / 8 limits the change in size
+  const markerScale = ((1 / zoom) + (Math.sin(Date.now() / 500) + 1) / 8) * extraScale;
+  const markerHeightHalf = 16 * markerScale;
+  const markerMarginAboveHealthBar = 10;
+
+  // TODO - move marker higher if the unit is a miniboss
```
jdoleary, line 487: You should be able to use unit.image.scale.y to determine this since minibosses and stacked summoned units are just scaled up.

src/graphics/PlanningView.ts
```diff
@@ -694,6 +626,31 @@ export async function runPredictions(underworld: Underworld) {
...
+      drawUICone(predictionGraphics, target, radius, startArc, endArc, colorOverride);
+    }
+    for (let { target, color, radius, text } of predictionCircles) {
+      const colorOverride = currentlyWarningOutOfRange ? colors.outOfRangeGrey : color;
+      drawUICircle(predictionGraphics, target, radius, colorOverride, text);
+    }
+  }
+  if (globalThis.radiusGraphics) {
+    for (let { target, color, radius, text } of predictionCirclesFill) {
+      //const colorOverride = currentlyWarningOutOfRange ? colors.outOfRangeGrey : color;
```
jdoleary, line 650: Prediction bloat is red instead of grey when enemy is out of range.  It should be grey to clearly denote to the user that this  _would_ be their bloat explosion radius but they're out of range and can't be cast on

jdoleary: I'm not sure if this is the area where that went wrong though, just noticed that it's commented out and might be an issue here too

src/graphics/ui/eventListeners.ts
```diff
@@ -851,6 +850,9 @@ export function clickHandler(overworld: Overworld, e: MouseEvent) {
       playSFXKey('deny');
     }
   } else {
+
+    const mouseTarget = underworld.getMousePos();
+    drawCircleUnderTarget(mouseTarget, underworld, 1.0, globalThis.planningViewGraphics);
```
jdoleary, line 855: What is this for? Click handler just triggers once on click and then is done.  Persistent circles on screen are usually drawn inside of recurring functions such as gameLoop or something that is called every frame

jdoleary (commented):
Looking really good! A few comments, questions and requested changes.

Also,
image
Attack range is labelled but bloat isn't.

jdoleary:
> IMPORTANT: The new unit event causes old saves to not load properly. This probably needs an exception handler or something. Let me know what I need to do for that, if anything 👍

To ensure backwards compatibility, all we have to do is supply a default for if the new object or array is undefined: 
for example,
`onDrawSelectedEvents: [...(onDrawSelectedEvents || [])],` This will spread the empty array if onDrawSelectedEvents is falsey.
I think there's 2 different places where this needs to be fixed.  The other is ...rest.onDrawSelectedEvents.

SoulMuncher:
> Attack range is labelled but bloat isn't.

There is only ever one "label text" at a time from the planning view, which includes the selectedunit and prediction visuals. I think this was something you did intentionally to prevent the screen getting too messy when you do something like [connect-circle-connect-circle...]

What's the expected behavior here? I can't decide if we should show everything or do something else

src/Underworld.ts
```diff
@@ -512,6 +503,7 @@ export default class Underworld {
         startPos.y += unitImageYOffset;
         const done = this.runForceMove(forceMoveInst, false);
         const endPos = { x: forceMoveInst.pushedObject.x, y: forceMoveInst.pushedObject.y + unitImageYOffset };
+        // TODO - Fix urn blood smear
```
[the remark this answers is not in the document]
SoulMuncher: This was fixed, I need to removed the TODO

src/entity/Unit.ts
SoulMuncher: Oh!... Probably not.

src/graphics/PlanningView.ts
SoulMuncher: Yeah pretty much! The label text previously had a font size of ~26 I think

src/graphics/PlanningView.ts
SoulMuncher: The background is dark grey currentlt. I think what you're seeing is the transparency, making it mix with color behind it.

I thought it looked a little strange too. Now I'm more confident it making it somewhat (or completely) opaque.

Also, you might've noticed non-mana units do not have a mana bar background. And if a non-mana unit does get mana, it will show as a full bar. I thought that was especially nice for send mana 🙂

src/graphics/PlanningView.ts
SoulMuncher: Another TODO I forgot to remove haha. Had to jump around a lot and do intermittent testing with these changes. Will clean it up shortly.

src/graphics/ui/eventListeners.ts
SoulMuncher: Um... I'm not really sure. In some situations where units and pickups were nearby, there was a discrepancy between what highlighted when you hovered, and what you actually selected when you clicked. I think I was testing something here and forgot to remove it.

jdoleary:
> > Attack range is labelled but bloat isn't.
> 
> There is only ever one "label text" at a time from the planning view, which includes the selectedunit and prediction visuals. I think this was something you did intentionally to prevent the screen getting too messy when you do something like [connect-circle-connect-circle...]
> 
> What's the expected behavior here? I can't decide if we should show everything or do something else

Yeah I did do this intentionally.  Most units are going to have an attack radius, so won't that mean that selecting them will always no show the "bloat explosion radius" label?  I think if the label isn't visible the circle shouldn't show because it's not obvious what it is then.

src/entity/Unit.ts
```diff
@@ -1452,25 +1452,23 @@ export function findLOSLocation(unit: IUnit, target: Vec2, underworld: Underworl
...
       }
     }
   }
 
-  if (!globalThis.selectedUnitGraphics) {
-    console.error("selectedUnitGraphics is undefined")
-    return;
-  }
+  // TODO - Ideally the logic below would be defined in each Unit's ts file individually
+  // Instead of using if/else and unit subtypes
```
jdoleary, line 1471: Why? If the logic is always the same for each given subtype wouldn't that just result in a lot of repetition for units that share the same subtype?

SoulMuncher: This would be for the same reason we put .action or .getUnitAttackTargets on each unit instead of in a large if/else. 

There are already many different ways a unit can draw information, and I imagine there will only be more as new units are added, especially when considering different kinds of doodads, aoe enemies, spellcasters, and so on. Many of the current drawings could be separated and improved too, such as the glops and priests. It would be nice to show more than just the attack range circle for most enemies, and I think that defining it in each enemy's script would be a cleaner way of doing this than to have the unit specific logic be in the broader unit class (and onDrawSelected is currently the only place this happens, afaik)

It also separates drawing behavior from unit.subtypes, making the subtypes more of an identification/targeting group thing, which I think is another step in the right direction.

SoulMuncher:
> Most units are going to have an attack radius, so won't that mean that selecting them will always no show the "bloat explosion radius" label? I think if the label isn't visible the circle shouldn't show because it's not obvious what it is then.

You're right that the bloat radius text is always hidden because the planning view is limited to one label currently. I don't think removing the circle is a good idea though. Like you said, the issue is that it's not obvious what the circle is, not that the circle exists in the first place. In a broader context, the whole point of the new onDraw system is to let players see more relevant information about the selected unit and to be able to make draw calls with modifiers. The problem/solution is finding a clean, clear, and concise way to convey that information and make those draw calls.

Even without any changes, I think that an unlabeled bloat circle is better than a non-existent one. If the player knows they put bloat on a unit, they'll figure out what the circle is, and if they don't know they put bloat on a unit, it's even more important that the circle is there to warn them imo. Especially when predictions don't consider ally moves.

Do you have any ideas for handling the label text? I could make a new issue out of it if it's helpful 👍

SoulMuncher:
> To ensure backwards compatibility

I might need some help with this. I am still running into issues on selection/bloat application for a v1.24 save. Think you can take a look at it?

src/entity/Unit.ts
jdoleary: Yeah that sounds good to me

jdoleary:
The new hp bar background looks great, and the trick with giving melee units a mana bar if they receive mana is brilliant.  Works exactly as i'd expect it to.
I'm going to merge this and I'll create a new issue to revisit the bloat label problem, but I want to get this one it since it was a ton of work!

jdoleary (approved):
Great work!

PR #164, opened 2023-12-16
Enemy Mana Adjustments

### LINKED ISSUE #162: Adjust unit max mana
I.E. Priest

Consider Mana Burn and Mana Steal

Do we want minibosses to have the same exact mana/regen as non minibosses?

SoulMuncher:
Closes #162
- Most units have less max mana, costs and regen adjusted slightly
- Mana scales when made miniboss

QA Single, Multi
I didn't have any save files with mana units present. Worth quickly checking load if you have a relevant save file.

.

Might need to buff Mana Burn potential/versatility in conjunction with these changes?
Maybe if a unit is out of Mana, it damages them based on regen or max mana or something. 

.

Note: Split still causes all priests, dark priests, summoners, and dark summoners to be unable to cast.

The solutions are
- Give units higher max mana values (buffs mana burn and steal, you can steal from one target a lot more, makes ui mana bars be half full at max, only fixes split 1)
- Give units cheaper casting costs (significant nerf to mana spells, similar ui issues)
- Change split mechanics (maybe doesn't affect max mana, just current and regen?)
- Let it be (might be a valid option to let the player disable the bigger mana-based units with split. my only complaint here would be about summ/dsumm/priest/dpriest **minibosses** specifically)


SoulMuncher:
Blood golem had 64 range where other melee units had 74. I went ahead and fixed this, as well as moved the base melee unit range to config

jdoleary:
I'm leaning towards this "Change split mechanics (maybe doesn't affect max mana, just current and regen?)" solution.  It doesn't exactly make sense, but I think that's okay, funtionality is more important and split would be way too powerful if it could just totally disable any caster.

src/entity/units/darkSummoner.ts
```diff
@@ -28,14 +28,14 @@ const unit: UnitSource = {
...
     attackRange: 550,
-    manaCostToCast,
-    manaMax: 60,
-    mana: 40,
-    manaPerTurn: 20,
-    bloodColor: 0x852124
+    healthMax: 180,
+    mana: 60,
+    manaMax: 120,
+    manaPerTurn: 15,
```
jdoleary, line 36: This means the dark summoner will only cast every 8 turns? Is this intentional?

src/entity/units/glop.ts
```diff
@@ -18,11 +18,14 @@ const unit: UnitSource = {
...
+    damage: 20,
     attackRange: 400,
-    bloodColor: bloodLobber,
     healthMax: 60,
-    damage: 20,
-    manaCostToCast: 15
+    mana: 30,
+    manaMax: 30,
+    manaPerTurn: 10,
+    manaCostToCast: 5,
```
jdoleary, line 27: 5 cost with 10 regen means the glop will never run out of mana on it's own.  Could be fine, just wondering why?

jdoleary (commented):
Looks good to me, 2 small questions

src/entity/units/darkSummoner.ts
SoulMuncher: Ah yeah... That's probably not frequent enough 😅

Will increase it a little!

src/entity/units/glop.ts
SoulMuncher: The glop is the first/most frequent mana enemy you see, so I sort've did this as a way to make mana burn/steal a little stronger. Both the regen and cost amounts felt right thematically / when compared to other units and their abilities.

jdoleary: Works for me!

SoulMuncher:
> Is this PR ready for review again?

Yes

jdoleary (approved):

jdoleary:
Great work!

Issue #166, opened 2023-12-16
HP bars not updating between stacked slash on golem


jdoleary:

SoulMuncher:
This happens when you queue up a spell while another one is still running, common in multiplayer.

The prediction is drawing "healing" values behind current health values because the prediction does not take into consideration the spell that is still going behind it.

Might be complicated to fix

jdoleary:
> Not sure how to check that for single/multi player

You could have a flag that gets set once a SPELL message starts processing and is cleared after it's done.

PR #183, opened 2023-12-17
Arrow Predictions

### LINKED ISSUE #63: Arrow Collision
Multiple exploding arrows shot from [(targetting spells) + Exploding Arrow] will repeatedly collide with the first enemy, even after that enemy has been killed. This leads to incorrect predictions.

SoulMuncher:
Closes #63

WIP - Visual discrepancies in multiplayer. #166 is very apparent here.

See latest comments

### DIFF
--- a/src/cards/arrow.ts
+++ b/src/cards/arrow.ts
@@ -10,6 +10,7 @@ import { add, equal, getAngleBetweenVec2s, getEndpointOfMagnitudeAlongVector, in
 import Underworld from '../Underworld';
 import { playDefaultSpellSFX } from './cardUtils';
 import { moveAlongVector, normalizedVector } from '../jmath/moveWithCollision';
+import { runPredictions } from '../graphics/PlanningView';
 
 export const arrowCardId = 'Arrow';
 const arrowProps: ArrowProps = {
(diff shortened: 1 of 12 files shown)


SoulMuncher:
As of the last commit [9d9f3f0]:

Arrow collisions are only calculated in predictions, which means predictions should be 100% accurate. This also means that arrows will keep the same visuals and coolness they had before and can still run synchronously without interfering with one another or other spell effects. Lastly, all arrows now use the basic arrowEffect function, which centralizes the code and makes it much more maintainable. There are three problems left to address:

- Shatter Arrow collides with initial target - Game loop only, probably a simple fix with the onCollide event promise.
- Stacked Exploding Arrow repeatedly hits target out of range - Prediction only, probably a simple fix with the onCollide event promise.
- All arrows look like they pierce in multiplayer. This is because predictedArrowCollisions[][] is not synced over the network. As far as I can tell, this is purely visual. I will do more testing once shatter/explosive arrow are fixed.

SoulMuncher:
Great idea @jdoleary 

One issue I'm seeing is that this inner prediction messes with the actual prediction. It's not entirely local. I'm not sure if it could be made local either since we need to run onCollide events and as an extention, things like forcePush.

jdoleary:
Can you reset the predictions with syncPredictionEntities between use?

SoulMuncher:
> @jdoleary Can you reset the predictions with syncPredictionEntities between use?

Sort've. It clears other spell predictions too, so it looks a little weird. If you were to just use slashes it would keep the prediction/death icon until the spell is done.

https://github.com/jdoleary/Spellmasons/assets/68413161/0c4caa71-44df-43ed-9c4b-0b7c1950e536

jdoleary:
Replaced by #303

PR #189, opened 2023-12-17
Player state source of truth

### LINKED ISSUE #188: client's copy of Player should be the source of truth
So that the client's player object is never overwritten from the server unless there's a LOAD_GAME_STATE.
The client will always have the most up to date copy of themself (read: `player`) and therefore, the client should be syncing the player object TO the server, not FROM the server.

Replaces #19 and #43

jdoleary:
for during game syncs.
The server is the source of truth when loading a game or when changing which player the client is controlling.

This should resolve the issue of player's losing Summon cards or weird behavior like having your upgrade options regenerate if your ally level's up while your choosing your upgrades.

Resolves #188

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -3808,11 +3808,11 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
   sendPlayerThinking(thoughts: { target?: Vec2, cardIds: string[] }) {
     sendPlayerThinkingThrottled(thoughts, this);
   }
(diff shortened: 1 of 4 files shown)


SoulMuncher:
I can't comment much on the code itself because I haven't worked with it much, but the idea of the client being the source of truth for certain things such as inventory sounds great! I don't foresee any major/unfixable issues with this approach.

SoulMuncher (approved):
Looks good!

PR #194, opened 2023-12-18
For i18n evaluation only

jdoleary:


src/cards/blood_curse.ts
```diff
@@ -24,29 +25,38 @@ function add(unit: IUnit, underworld: Underworld) {
...
-    }
   });
 
+  if (!prediction) {
+    updateTooltip(unit);
+  }
+
+  if (unit.modifiers[suffocateCardId]) {
+    updateSuffocate(unit, underworld, prediction);
+  }
```
jdoleary, line 42: @SoulMuncher Maybe this is okay for a one-off, but seeing logic for suffocate in another spell's file seems like a footgun to me.  Maybe there's a better way?
Is it because the max health is changing and so suffocate needs to update?

jdoleary: Note: This PR is just for me to evaluate what localization tasks I have but it was convenient for me to leave this comment for you while I was looking through it

SoulMuncher: jd
> Maybe this is okay for a one-off, but seeing logic for suffocate in another spell's file seems like a footgun to me. Maybe there's a better way?
Is it because the max health is changing and so suffocate needs to update?

Yes to all of that, which is why I brought it up in discord the other day:

Past me on discord
> If suffocate buildup > current hp, unit dies.
Assume buildup is 80, and current health is 90
If I deal 20 damage, the unit dies instantly, and it does this by checking for the modifier in Unit.TakeDamage. Even in another situation where I don't kill it, I need to update the turns until death tooltip. This is fine for most cases, but then what if I split the unit, or cast blood curse on it?
I'm now handling all of these cases manually by checking for the modifier in that location/area of code, but in the long term that's probably not the ideal solution. I'm not sure how to go about this... With a solid stat system you would call some onStatModified event that would then update suffocate, but we don't have that. 
I think it would take a big refactor to come up with a good long-term solution. It's mostly a code cleanliness and future bug-proofing thing anyway, but I thought I'd mention it to you in case you have any good ideas. This may have to do for 1.26

jdoleary:
1.26

PR #196, opened 2023-12-19
Additional Balancing and Fixes (Save for 1.27+)

### LINKED ISSUE #168: Level spawn budget issue
Level 5 only had 1 priest and 2 glops

SoulMuncher:
Some balancing and bug fixes

- Heal Allies mana cost increased from 25 to 40, heal increased from 10 to 40
- Shield mana cost decreased from 60 to 40, expense scaling decreased from 3 to 2, cooldown removed
- Decoy HP increased from 70 to 80
- Slow amount increased from 20% to 25%
- Arrow 3 mana cost decreased from 22 to 20, damage increased from 30 to 40
- Send Mana mana cost decreased from 30 to 20
- Contaminate mana cost decreased from 50 to 40
- Debilitate mana cost increased from 20 to 30, scaling decreased from 3 to 2
- Resurrect mana cost decreased from 160 to 120, cooldown decreased from 2 to 1
- Weak Resurrect mana cost decreased from 70 to 60
- Toxic Resurrect cooldown decreased from 2 to 1
- Clone mana cost increased from 80 to 120, expense scaling reduced from 4 to 2
.
- Increased trap damage from 30 to 40
- Increased difficulty and variety of midgame (enemy types appear faster, and increased midgame enemy budget)
- Fixed a bug that sometimes caused most of a level's budget to be taken up by one unit type
- Fixed a bug that caused only a couple enemies to spawn in one level
- Reworked budget and probability values to create more interesting maps and a better difficulty curve for the new stats
- Rebalanced some units

Moved some changes to #204
- Removed suffocate from starting damage spells
- Cast range is displayed as full cast range instead of being divided by 2
- Fixed a bug that caused decoy to get SUPER massive
- Fixed mana cost to cast tooltip issue for minibosses

Closes #168


SoulMuncher:
Unit budget changes may have changed some summon_generic costs. I need to look at that before we pull this

jdoleary:
I appreciate the detailed change log, in the future can you please also include reasoning as to why balancing changes were made?

src/Underworld.ts
```diff
@@ -4134,7 +4130,7 @@ function getEnemiesForAltitude(underworld: Underworld, levelIndex: number): stri
   // Prevent negative values which can happen during tutorial
   const adjustedLevelIndex = Math.max(0, levelIndex);
 
-  const numberOfTypesOfEnemies = 2 + Math.floor(adjustedLevelIndex / 2);
+  const numberOfTypesOfEnemies = Math.min(2 + Math.floor(adjustedLevelIndex / 2), 6);
```
jdoleary, line 4133: This would mean that the numberOfTypesOfEnemies never goes above 6.
What's the reasoning here?

src/Underworld.ts
```diff
@@ -2778,10 +2778,6 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
                 }
                 return '';
               }
-              if (stat == 'attackRange') {
-                // Display default attackRange as "100"
-                return Math.floor(100 * value / config.PLAYER_BASE_ATTACK_RANGE);
-              }
```
jdoleary, line 2784: Why was this removed?

src/cards/arrow3.ts
```diff
@@ -5,14 +5,14 @@ import { arrowEffect } from './arrow';
 import { arrow2CardId } from './arrow2';
 
 export const arrow3CardId = 'Arrow3';
-const damageDone = 30;
+const damageDone = 40;
```
jdoleary, line 8: This is a pretty significant jump. Why this change?

src/cards/bleed.ts
```diff
@@ -36,7 +36,7 @@ const spell: Spell = {
     manaCost: 10,
     healthCost: 0,
     expenseScaling: 1,
-    probability: probabilityMap[CardRarity.RARE],
+    probability: probabilityMap[CardRarity.UNCOMMON],
```
jdoleary, line 39: Bleed is a fairly potent spell, why is it made more common?  This makes it 3x more likely to appear

src/cards/bloat.ts
```diff
@@ -64,7 +64,7 @@ const spell: Spell = {
     manaCost: 15,
     healthCost: 0,
     expenseScaling: 1,
-    probability: probabilityMap[CardRarity.COMMON],
+    probability: probabilityMap[CardRarity.UNCOMMON],
```
jdoleary, line 67: Bloat is one of the funnest spells and isn't super powerful (it doesn't scale well to late game).  I'm comfortable with it being COMMON or SPECIAL but UNCOMMON seems too rare especially because it's hardly useful late game

src/cards/debilitate.ts
```diff
@@ -16,9 +16,9 @@ const spell: Spell = {
     category: CardCategory.Curses,
     sfx: 'debilitate',
     supportQuantity: true,
-    manaCost: 20,
+    manaCost: 30,
```
jdoleary, line 19: Good change 👍

src/cards/heal_mass.ts
```diff
@@ -8,7 +8,7 @@ import { CardRarity, probabilityMap } from '../types/commonTypes';
 import { heal_greater_id } from './heal_greater';
 
 export const heal_mass_id = 'Mass Heal';
-const healAmount = 10;
+const healAmount = 40;
```
jdoleary, line 11: So regular heal gives 2hp to 1 mana ratio and this does 1hp to 1mana ratio, but it heals all allies.  I agree that the heal amount should be increased, 10 is pretty weak but maybe it should cost a little more?

src/cards/lastwill.ts
```diff
@@ -35,7 +35,7 @@ const spell: Spell = {
     manaCost: 15,
     healthCost: 0,
     expenseScaling: 1,
-    probability: probabilityMap[CardRarity.COMMON],
+    probability: probabilityMap[CardRarity.UNCOMMON],
```
jdoleary, line 38: Why the change here?  Last Will is pretty key for early - mid game scaling.

src/cards/poison.ts
```diff
@@ -63,7 +63,7 @@ const spell: Spell = {
     manaCost: 20,
     healthCost: 0,
     expenseScaling: 1,
-    probability: probabilityMap[CardRarity.UNCOMMON],
+    probability: probabilityMap[CardRarity.COMMON],
```
jdoleary, line 66: 👍

src/cards/resurrect.ts
```diff
@@ -17,9 +17,9 @@ const spell: Spell = {
     replaces: [resurrect_weak_id],
     category: CardCategory.Soul,
     sfx: 'resurrect',
-    manaCost: 160,
+    manaCost: 120,
```
jdoleary, line 20: reasoning?

src/cards/resurrect_toxic.ts
```diff
@@ -20,7 +20,7 @@ const spell: Spell = {
     supportQuantity: true,
     manaCost: 90,
     healthCost: 0,
-    cooldown: 2,
+    cooldown: 1,
```
jdoleary, line 23: 👍

src/cards/resurrect_weak.ts
```diff
@@ -16,7 +16,7 @@ const spell: Spell = {
     id: resurrect_weak_id,
     category: CardCategory.Soul,
     sfx: 'resurrect',
-    manaCost: 70,
+    manaCost: 60,
```
jdoleary, line 19: 👍

src/cards/send_mana.ts
```diff
@@ -21,10 +21,10 @@ const spell: Spell = {
     category: CardCategory.Mana,
     sfx: 'potionPickupMana',
     supportQuantity: true,
-    manaCost: 30,
+    manaCost: 20,
```
jdoleary, line 24: This should be 30.  We talked about this at length, why the change?  It's super broken (OP with timemason) if the send to receive ratio is 1:1

src/cards/shield.ts
```diff
@@ -18,11 +18,10 @@ const spell: Spell = {
     category: CardCategory.Blessings,
     sfx: 'shield',
     supportQuantity: true,
-    manaCost: 60,
+    manaCost: 40,
```
jdoleary, line 21: 👍 Good change, it makes shield a lot more viable

src/cards/summon_decoy.ts
```diff
@@ -62,11 +62,6 @@ const spell: Spell = {
           // Animate effect of unit spawning from the sky
           skyBeam(unit);
         }
-        if (unit.image) {
-          const quantityScaleModifier = 1 + 0.3 * (quantity - 1);
-          unit.image.sprite.scale.x = unit.image.sprite.scale.x * quantityScaleModifier;
-          unit.image.sprite.scale.y = unit.image.sprite.scale.y * quantityScaleModifier;
-        }
```
jdoleary, line 69: Is this already handled by the other summon logic?

src/entity/Pickup.ts
```diff
@@ -495,7 +495,7 @@ export function tryTriggerPickup(pickup: IPickup, unit: IUnit, underworld: Under
 
 const manaPotionRestoreAmount = 40;
 const healthPotionRestoreAmount = 50;
-export const spike_damage = 30;
+export const spike_damage = 40;
```
jdoleary, line 498: why?

src/entity/units/ancient.ts
```diff
@@ -16,7 +16,7 @@ const unit: UnitSource = {
     subtype: UnitSubType.RANGED_RADIUS,
   },
   unitProps: {
-    damage: 3,
+    damage: 10,
```
jdoleary, line 19: My intent is that the ancient is a small, persistent danger.  If you don't deal with them they'll become a problem but aren't an immediate threat.  With the higher base health, I can understand this mostly.  Can you provide more context as to why this change?

src/entity/units/archer.ts
```diff
@@ -17,14 +17,14 @@ const unit: UnitSource = {
...
+    damage: 20,
+    attackRange: 400,
+    healthMax: 20,
     manaMax: 0,
   },
   spawnParams: {
-    probability: 50,
-    budgetCost: 2,
+    probability: 10,
+    budgetCost: 3,
```
jdoleary, line 27: I could be talked into this.   Double damage is a lot, but I guess golems have a pretty far stamina range so maybe this makes sense? I'd like to hear your reasoning

jdoleary: That's a significantly less probability

src/entity/units/bloodGolem.ts
```diff
@@ -16,15 +16,15 @@ const unit: UnitSource = {
   },
   unitProps: {
     damage: 40,
-    staminaMax: config.UNIT_BASE_STAMINA * 1.2,
-    healthMax: 60,
+    staminaMax: config.UNIT_BASE_STAMINA,
+    healthMax: 80,
```
jdoleary, line 20: Keep in mind, the stamina of enemies relative to the player's cast range significantly determines their threat.  If the player has to step into danger in order to cast on them, that's a big part of the player's calculus.

src/entity/units/bloodGolem.ts
```diff
@@ -16,15 +16,15 @@ const unit: UnitSource = {
...
     manaMax: 0,
     bloodColor: 0x8a2e2e,
   },
   spawnParams: {
     probability: 100,
-    budgetCost: 4,
-    unavailableUntilLevelIndex: 4,
+    budgetCost: 10,
+    unavailableUntilLevelIndex: 3,
```
jdoleary, line 27: This feels too early to show an upgraded version of the regular golem

src/entity/units/darkPriest.ts
```diff
@@ -19,17 +19,17 @@ const unit: UnitSource = {
...
     manaPerTurn: 30,
     manaCostToCast: 60,
   },
   spawnParams: {
-    probability: 20,
-    budgetCost: 10,
-    unavailableUntilLevelIndex: 8,
+    probability: 40,
+    budgetCost: 15,
+    unavailableUntilLevelIndex: 6,
```
jdoleary, line 32: This is really early for a multi-cast enemy

src/entity/units/golem.ts
```diff
@@ -14,14 +14,14 @@ const unit: UnitSource = {
...
-    healthMax: 20,
+    damage: 20,
+    staminaMax: config.PLAYER_BASE_ATTACK_RANGE,
+    healthMax: 40,
     manaMax: 0,
   },
   spawnParams: {
-    probability: 100,
-    budgetCost: 1,
+    probability: 10,
```
jdoleary, line 23: 90% reduction in probability for the core basic unit??

src/entity/units/golem.ts
```diff
@@ -14,14 +14,14 @@ const unit: UnitSource = {
     subtype: UnitSubType.MELEE,
   },
   unitProps: {
-    damage: 30,
-    staminaMax: config.UNIT_BASE_STAMINA,
-    healthMax: 20,
+    damage: 20,
```
jdoleary, line 17: Same damage as the archer?

src/entity/units/gripthulu.ts
```diff
@@ -28,8 +30,8 @@ const unit: UnitSource = {
   },
   spawnParams: {
     probability: 20,
-    budgetCost: 4,
-    unavailableUntilLevelIndex: 7,
+    budgetCost: 12,
+    unavailableUntilLevelIndex: 6,
```
jdoleary, line 34: I don't think gripthulus are enough of a threat to take up this much budget

jdoleary (changes requested):
This is a huge amount of last minute balancing changes.  Let's save this for the next update instead of 1.26.  Some of these changes I can totally get behind and some I am very confused by.  Let's talk about them.

src/Underworld.ts
SoulMuncher: Beyond 6 enemy types means the levels lose their theme.

Now that I'm thinking about, I don't think this even comes into play until post deathmason. Happy to revert

src/Underworld.ts
SoulMuncher: I think this makes sense in the context of the new stat values, essentially it's a lot easier to go over 200 displayed cast range, and at that point having the display divided by 2 may just be more confusing / more calculation. You don't actually know what your cast range is relative to stamina and enemy attack ranges because the displayed value is arbitrarily divided by 2.

I did see the feedback on discord from a while back saying that "240 cast range doesn't mean anything to me", but they were asking for range in terms of "tiles", and the current 1/2 display doesn't really solve that feedback, it just makes the number smaller. At least now it would have some meaning/reference in the game world

src/cards/arrow3.ts
SoulMuncher: Slash goes 20-40-80, this makes arrow go 10-20-40.

Also, Arrows are very bad due to having low damage. You can't reliably use it as a primary damage spell because you will have to cast it twice as much as slash, and this means the cost will scale significantly, and wont come back for turns at a time. Archer solves this problem, but I'd like the spell to be usable for all classes

src/cards/bleed.ts
SoulMuncher: I did this because bleed is fun and it feels like I never see it/get to use it.

I don't feel too strongly about this change, just thought it was nice to give more non-slash options. It is a very powerful spell for "UNCOMMON", to be fair.

src/Underworld.ts
jdoleary: Sounds good to me, we can leave it in.

src/Underworld.ts
jdoleary: Works for me!

src/cards/arrow3.ts
jdoleary: Hmm, well arrows do have the advantage of infinite range.  We'll have to play test it a bit and see.  I see your logic here

src/cards/bleed.ts
jdoleary: I feel like it's a pretty critical end game build spell.  I think we should leave it Rare, simply because it's so potent and it occurring less frequently will entice people to take it when they see it and use it

src/cards/bloat.ts
SoulMuncher: Bloat doesn't seem like a "core" spell that should go along with stuff like slash and push, and I see it way more often than I'd like. It's not entirely applicable to most builds, and feels like the mechanics are special/interesting enough to warrant uncommon.

Maybe my real concern with bloat is how weak/messy it is in late game. It's generally the curse I want to have the least out of all of them, so I avoid picking it and maybe thats why it seems so common to me.

src/cards/heal_mass.ts
SoulMuncher: I did also increase the cost to 40. If you mean more than that... maybe?

Given it's a 3rd tier upgrade forbidden rarity spell, I don't mind making it a little more powerful than not

src/cards/resurrect.ts
SoulMuncher: The real resurrect feels like a worse version of the other spells, since you can heal or purify or whatever else.

This change doesn't fix that (it shouldn't, players should be rewarded for finding those combos) but it does at least make the full resurrect feel competitive, which it should, given it's supposed to be an upgrade. Right now you kind've have to use one of the other combos instead

src/cards/send_mana.ts
SoulMuncher: I know, but I do still really think the the expense scaling fixes that.

I gave the 30 base cost a good try and I just don't think it's the play. It feels too expensive to use in those clutch moments we talked about, and it feels too expensive/niche to ever choose over other spells when picking upgrades.

src/cards/summon_decoy.ts
SoulMuncher: This is handled in unit.create now, I think.

I'd have to check, but yes it's handled elsewhere

src/entity/units/ancient.ts
SoulMuncher: It's primarily due to increased max health. They also cast a bit less due to mana changes

src/entity/units/archer.ts
SoulMuncher: Archers deal practically 0 damage. In my last test with Luna we both felt we could just ignore them entirely, and it became a chore to chew through them since they weren't threats to us.

I think it also makes sense for ranged guys to be squishy and golems to be tankier etc.

Other Archers are WAYYY scarier, despite similar budget costs.

And yes I noticed that about the probability. This makes it more likely to choose interesting units in later levels, rather than mass default-archer

src/entity/units/bloodGolem.ts
SoulMuncher: It feels like it's time to ramp up once you get to the lava levels. HP changes make midgame kind've easy, and this is supposed to help with that

src/entity/units/darkPriest.ts
SoulMuncher: You think so? Maybe. I've never had any struggles with dark priests, and I figured increased enemy variety would be good

If you're saying it's too strong maybe we should increase the budget cost instead?

Again though: Midgame is too easy right now. Really the whole game, but especially midgame.

src/entity/units/golem.ts
SoulMuncher: Yes, the goal is less cannon fodder in later levels. It's more likely to pick non-basic golem/archer/glop.

src/entity/units/golem.ts
SoulMuncher: Yes. Do you feel that's too low/high?

The basic golems are meant to be a little weaker in this version

src/entity/units/gripthulu.ts
SoulMuncher: 12 is on the lower end now: a blood golem/archer only takes 10

We could lower it, this was mainly to distringuish it from filler units.

It also stems from the enemy being.. kind've annoying. Lol. It might not kill you, but it's very intrusive by naturr and thus I thought it should have slightly higher budget than the other guys. I think it was buffed here too.

src/cards/bleed.ts
SoulMuncher: That makes sense!

jdoleary:
We can revisit balancing a bit at a time, but for now I'm closing this since it is very large and the impact of changing so much at once is undesirable to me.

Issue #236, opened 2023-12-20
End Turn button should say "Ready" on multiplayer instead of end turn


jdoleary:
Needs localization

jdoleary:
Credit: MrMarblz

Couls:
This can be closed now

jdoleary:
Closed via #240. Thank you @Couls !

PR #240, opened 2023-12-20
Change end turn to ready in multiplayer

Couls:
Implements #236 with i18n support, Ready already exists in localization, dynamically changes it when playing multiplayer and singleplayer

### DIFF
--- a/src/network/networkHandler.ts
+++ b/src/network/networkHandler.ts
@@ -37,6 +37,7 @@ import { setPlayerNameUI } from '../PlayerUtils';
 import { GameMode } from '../types/commonTypes';
 import { recalcPositionForCards } from '../graphics/ui/CardUI';
 import { isSinglePlayer } from './wsPieSetup';
+import { elEndTurnBtn } from '../HTMLElements';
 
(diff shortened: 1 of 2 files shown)


src/network/networkHandler.ts
```diff
@@ -682,6 +683,10 @@ async function handleOnDataMessage(d: OnDataArgs, overworld: Overworld): Promise
               // that they have, run the cinematic again.
               runCinematicLevelCamera(underworld);
             }
+            // Change end turn button from End Turn to Ready in multiplayer
+            const elEndTurnSpan = elEndTurnBtn.querySelector('[data-localize-text]') as HTMLElement;
+            elEndTurnSpan.dataset.localizeText = "ready"
```
Couls, line 688: This line could probably be removed and we can directly just do elEndTurnSpan.innerText = i18n("ready") for example but I opted to keep it in to keep the HTML elements consistent

jdoleary: localizeText automatically re-localizes the text when the games language is changed, so this is actually necessary in addition to setting the innerText

jdoleary (approved):
Great work! You're crushing it with these PRs :) Thank you!

PR #260, opened 2023-12-21
1.26 bug fixes 1

### LINKED ISSUE #72: The unit stats on miniboss summon cards show the stats for a non miniboss unit.
Ex: golem minibosses have 60 health and do 60 damage but the card shows the regular golem stats

### LINKED ISSUE #213: suffocate tooltip typo with  the ","

### LINKED ISSUE #250: Replaced spells in tooltip
Replaced spells are not removed from spellmason/player tooltip spell list

### LINKED ISSUE #251: Dark Priest prediction
- Enemy dark priest didn't predict to attack despite having 60 mana and players in range
- Enemy dark priest miniboss was split with 90 mana, predicted no attack, and then attacked

Likely reading max mana instead of the unit's mana-cost-to-cast or something

### LINKED ISSUE #252: Health cost badge while shielded
Mana Steal and similar health-costing spells have their badge cost reduced by your shield amount, and can even disappear entirely.

I would expect it to still show the cost of the spell, rather than showing my final health depletion. This applies to fortify as well. Reasoning: I can already look at my health bar to see my final health, I can't look anywhere to see whether mana steal is taking 47 or if it's already reset back to the 30 base cost, except the badge. It's functionally the same as having overfilled mana, but we don't hide mana badges when that happens, so we shouldn't hide health either imo

### LINKED ISSUE #253: Split + Purify doesn't restore mana regen
Should make the unit's mana regen go back to normal. Noticed this on a dark summoner miniboss that I killed and then revived.

### LINKED ISSUE #254: Purify Target Corpse
Purify can't affect corpses, and doesn't get refunded when you cast it on them.

Noticed this when trying to remove split from a dead enemy. Could have more significant impact in the future with effects like "undying"

SoulMuncher:
Closes #72 
- Summon stats now account for miniboss modifiers

Closes #250
- Replaced spells no longer appear in the player tooltip

Closes #251
- Fixed an issue with Dark Priest attack prediction

Closes #252
- Spells display their unadjusted health cost while shielded/fortified

Closes #253
- Mana Regen is restored upon split removal

Closes #254
- Purify can now target corpses

Closes #213
- Fixed suffocate tooltip

Also fixed an issue that caused some cards to say "upgrade" when they actually only "require" the following pre-requisite.

### DIFF
--- a/src/cards/suffocate.ts
+++ b/src/cards/suffocate.ts
@@ -84,7 +84,7 @@ export function updateTooltip(unit: Unit.IUnit) {
     const turnsUntilSuffocation = Math.ceil(2 * Math.log2(unit.health / 10) + 1) - modifier.quantity;
 
     // Set tooltip:
-    modifier.tooltip = i18n(['turns until suffocation', turnsUntilSuffocation.toString()]);
+    modifier.tooltip = `${turnsUntilSuffocation} ${i18n('turns until suffocation')}`;
   }
 }
 
--- a/src/cards/summon_generic.ts
+++ b/src/cards/summon_generic.ts
@@ -27,181 +27,195 @@ import { DARK_PRIEST_ID } from '../entity/units/darkPriest';
 
 
 const overrides: { [unitId: string]: { exclude: boolean, properties: { manaCost?: number } } } = {
-    'decoy': {
-        exclude: true,
-        properties: {}
-    },
-    [golem_unit_id]: {
-        exclude: false,
-        properties: { manaCost: 60 }
-    },
-    [ARCHER_ID]: {
-        exclude: false,
-        properties: { manaCost: 60 }
-    },
-    [ANCIENT_UNIT_ID]: {
-        exclude: false,
-        properties: { manaCost: 50 }
-    },
-    [GLOP_UNIT_ID]: {
-        exclude: false,
-        properties: { manaCost: 80 }
-    },
-    [gripthulu_id]: {
-        exclude: false,
-        properties: { manaCost: 80 }
-    },
-    [BLOOD_GOLEM_ID]: {
-        exclude: false,
-        properties: { manaCost: 100 }
-    },
-    [POISONER_ID]: {
-        exclude: false,
-        properties: { manaCost: 100 }
-    },
-    [VAMPIRE_ID]: {
-        exclude: false,
-        properties: { manaCost: 150 }
-    },
-    [BLOOD_ARCHER_ID]: {
-        exclude: false,
-        properties: { manaCost: 130 }
-    },
-    [PRIEST_ID]: {
-        exclude: false,
-        properties: { manaCost: 130 }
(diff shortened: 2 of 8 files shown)


src/cards/suffocate.ts
```diff
@@ -84,7 +84,7 @@ export function updateTooltip(unit: Unit.IUnit) {
     const turnsUntilSuffocation = Math.ceil(2 * Math.log2(unit.health / 10) + 1) - modifier.quantity;
 
     // Set tooltip:
-    modifier.tooltip = i18n(['turns until suffocation', turnsUntilSuffocation.toString()]);
+    modifier.tooltip = `${turnsUntilSuffocation} ${i18n('turns until suffocation')}`;
```
jdoleary, line 87: This should've worked as is, if it didn't there may be something wrong with the i18n function.  It's supposed to replace the first bread emoji and there's one at the start of the localization string.

src/cards/summon_generic.ts
```diff
@@ -27,181 +27,195 @@ import { DARK_PRIEST_ID } from '../entity/units/darkPriest';
...
+  }
 
-    const expenseScaling = 2;
-    const manaCost = override?.properties.manaCost ? (override.properties.manaCost * (asMiniboss ? 1.5 : 1)) : Math.max(1, Math.round(2 * Math.log2((sourceUnit.spawnParams?.budgetCost || 1)))) * 30 * (asMiniboss ? 2 : 1);
-    const id = unitId + (asMiniboss ? ' Miniboss' : '');
-    const unitSource = allUnits[unitId];
-    let healthMax = unitSource?.unitProps.healthMax || config.UNIT_BASE_HEALTH;
-    let manaMax = unitSource?.unitProps.manaMax || 0;
+  const expenseScaling = 2;
+  const manaCost = override?.properties.manaCost ? (override.properties.manaCost * (asMiniboss ? 1.5 : 1)) : Math.max(1, Math.round(2 * Math.log2((sourceUnit.spawnParams?.budgetCost || 1)))) * 30 * (asMiniboss ? 1.5 : 1);
```
jdoleary, line 128: I think there's an override for every single summon now.  So this line shouldnt do anything other than use the override.  Why was this changed?

jdoleary (commented):
Looks great overall, a few questions for clarification

src/cards/summon_generic.ts
SoulMuncher: Green glop was missing an override.

I also thought we'd want the miniboss multiplier to be the same for both overwritten and budget based summons

src/cards/suffocate.ts
SoulMuncher: Oh interesting. The format it's using now is similar to poison, but maybe poison is a weird exception because it has multiple variables.

Should we look at the i18n function?

jdoleary (approved):

PR #270, opened 2023-12-21
Draft for hotfix

jdoleary:


src/cards/split.ts
```diff
@@ -49,6 +49,7 @@ function remove(unit: Unit.IUnit, underworld: Underworld) {
   unit.manaMax = manaMax;
   // Prevent unexpected overflow
   unit.mana = Math.min(manaMax, unit.mana);
+  unit.manaPerTurn = manaPerTurn;
```
jdoleary, line 52: @SoulMuncher Do we want manaPerTurn cut in half when mana isn't? It'll make all the splits very slow to recover their mana.

src/cards/summon_generic.ts
```diff
@@ -27,181 +27,195 @@ import { DARK_PRIEST_ID } from '../entity/units/darkPriest';
...
+  [PRIEST_ID]: {
+    exclude: false,
+    properties: { manaCost: 130 }
+  },
+  [SUMMONER_ID]: {
+    exclude: false,
+    properties: { manaCost: 150 }
+  },
+  [GHOST_ARCHER_ID]: {
+    exclude: false,
```
jdoleary, line 79: Didn't you mention that you added green glop here? Its still missing

jdoleary: @SoulMuncher

SoulMuncher: Sorry for the conufsion, I did not add it here, I only changed the miniboss scaling to be the same for overwritten units and those unaccounted for

src/cards/split.ts
SoulMuncher: @jdoleary I thought mana was cut in half as well, so I'm not quite sure what you're asking.

Mana per turn is cut in half so that if you split a summoner or priest, it doesn't make them stronger by having two of them casting the spell at the normal speed. Cutting mana per turn was the solution we came up with to splitting a summoner and then the splits being completely unable to cast anything due to lower max mana than casting cost.

jdoleary: Oh I see the costToCast remains the same and so does the max mana.
This might need to still be balanced, this means a split will make a summoner take 2x times as long to recharge mana, which is a long time.  Maybe that's the benefit of split,  but it feels like it might be OP for casters

PR #278, opened 2023-12-22
Send Mana Id (i18n)

### LINKED ISSUE #224: i18n: Missing translation "send_mana"

SoulMuncher:
Closes #224 

Upgrade.ts was searching for localization id "send_mana" but in the sheet it is called "Send Mana". I went ahead and just changed the spell id to match the localization.

@jdoleary 
There doesn't seem to be any standard naming convention for spell ids. Do you want me to clean this up? Also, should we sort the localization sheet alphabetically to group similar things together/prevent duplicates? It would only take a few clicks, but I wasn't sure if it was left the way it is for a specific reason.


jdoleary (approved):
I'm going to try to build a localization tool over Christmas break, so no need to organize the sheet.
Also I think changing the ids would interfere with save files unfortunately.  But we could determine a standard for future ones

PR #279, opened 2023-12-22
Polished Power Bar

SoulMuncher:
- Fixed an issue that caused level up to give too much xp
- Fixed an issue that caused options to not get updated when opening the power bar, and caused an admin command different than what was selected to run
- Your previous selection will stay even after closing the bar, letting you quickly run the command several times (good for level up, regenerate level, spawning enemies, etc.)
- Tab will now increment the selected index
- Selected index will now reset when you start typing (since the selected command will likely change anyway)


jdoleary (approved):
Looks great!

PR #294, opened 2023-12-26
EndTurnBtn cannot be undefined

SoulMuncher:
Getting a few errors that this may be undefined and blocking multiplayer testing.

Not sure if it's ever actually undefined, hope this is a good fix 👍

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -2373,25 +2373,23 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
   // Sends a network message to end turn
   async endMyTurnButtonHandler() {
     if (globalThis.player) {
-      if (elEndTurnBtn) {
-        // If End turn button is level up button
-        if (elEndTurnBtn.classList.contains('upgrade') || document.body?.classList.contains(showUpgradesClassName)) {
-          const upgradesLeftToChoose = this.upgradesLeftToChoose(globalThis.player)
-          if (upgradesLeftToChoose > 0) {
-            this.showUpgrades();
-            elEndTurnBtn.classList.toggle('upgrade', false);
-            // Do not end turn, just show upgrades
-            return;
-          } else {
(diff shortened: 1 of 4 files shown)


jdoleary (changes requested):
So it actually needs to be typed as `| undefined` because in the headless context it is always undefined and will throw an error on headless if it is ever used.  This typing makes it so typescript enforces checking that it exists as not undefined before use.
This PR should instead add `if` checks around the uses of elEndTurnBtn where the errors are and leave the `| undefined` type.
In fact, all elements in this file should contain `| undefined`

SoulMuncher:
Thanks for the comment! That makes sense

SoulMuncher:
Should be good to go now 😄

src/Underworld.ts
```diff
@@ -2373,23 +2373,25 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
...
-      // If End turn button is level up button
-      if (elEndTurnBtn.classList.contains('upgrade') || document.body?.classList.contains(showUpgradesClassName)) {
-        const upgradesLeftToChoose = this.upgradesLeftToChoose(globalThis.player)
-        if (upgradesLeftToChoose > 0) {
-          this.showUpgrades();
-          elEndTurnBtn.classList.toggle('upgrade', false);
-          // Do not end turn, just show upgrades
+      if (elEndTurnBtn) {
+        // If End turn button is level up button
+        if (elEndTurnBtn.classList.contains('upgrade') || document.body?.classList.contains(showUpgradesClassName)) {
```
jdoleary, line 2378: small nitpick: This could be rewritten as `if(elEndTurnBtn && elEndTurnBtn.classList...` to reduce the nested if statements

jdoleary (approved):
Small nitpick, but I'll approve regardless.

src/Underworld.ts
SoulMuncher: No problem, happy to make it look cleaner. This is fixed in 44b92cb

jdoleary (approved):

PR #296, opened 2023-12-27
* Added ignoreRange flag to ICard

mattthewaz:
* Added ignoreRange flag to ICard
* Set to true for arrow spells
* Updated check in isAllowedToCastOutOfRange to use ignoreRange

This should allow new cards to be added that can also ignore range restrictions without having to include 'arrow' in that card's id.

This will need an accompanying PR into the mods repo to update the ICard interface, so that it also includes ignoreRange.


jdoleary (approved):

jdoleary:
Great work!

PR #303, opened 2023-12-28
Refactor Force move projectile

### LINKED ISSUE #63: Arrow Collision

jdoleary:
**This changes refactors projectiles to be a part of the forceMove loops instead of running on setTimeout.**

This change is intended to fix issues that result from headless and prediction running different simulations from the core game loop.

In it's current state this PR already fixes the many arrows on bloated unit desyncs that have been reported; however, it can go a bit further still.
Eventually I would like setTimeout to be simulated inside of headless and prediction so that it simulates waiting (relative to other code that takes time to execute (deltaTime'd gameloop) but will still execute as soon as possible.
(headlessTimeouts on underworld is in progress)

Here is an example of how this system is MUCH improved over the previous:
In this picture, the middle archer has an artificial 200 health, the first arrow explodes on the first archer as expected and the second arrow now correctly explodes on the 2nd archer.  The only imperfect outcome is that there is no second push on the second archer from the second explosion.
image
Resolves #63


SoulMuncher:
@jdoleary 
Phantom arrows can hit the same target multiple times

SoulMuncher:
There's some weird behavior that happens when you're aiming at a corner. The arrow path changes and doesn't go straight through the cursor anymore.

SoulMuncher:
Bloat predictions are still incorrect:

https://github.com/jdoleary/Spellmasons/assets/68413161/2336a3c4-10ea-414c-81b5-03600d3b399b

SoulMuncher:
Impact damage is a little broken, you can deal upwards of 350 damage with one shove. Two shoves kills the deathmason lol.

Impact damage should probably use some kind of sqrt or log-based calculation.

@jdoleary
ALSO: When calculating impact damage based on "true magnitude", shove debugs to ~800 in prediction, and usually less than half that in the actual cast, which makes me think the prediction is still doing the force moves all at once, without simulating the time between them and decreasing the velocity each loop or something. Didn't look too closely though

SoulMuncher:
You can push things out of bounds via pushing them into the corners of walls: (Both in prediction and actual game loop)

CornerPush1

SoulMuncher:
Last arrow isn't awaited: Just enough arrows to kill a unit + blood curse will cause the blood curse to trigger before the last arrow hits, and the unit won't actually die as predicted.

jdoleary:
Still todo: Impact damage is different in prediction vs real gameloop ( i think this will be fixed with future parity improvements)
can't reproduce: Pushing units through corners.
The rest has been fixed

SoulMuncher:
> There's some weird behavior that happens when you're aiming at a corner. The arrow path changes and doesn't go straight through the cursor anymore.

I'm still seeing weird arrow behavior when aiming around walls/corners. It also "jumps" when aiming along a flat wall. Do objects stop in place before colliding with a wall, instead of moving all the way to it, like a partial/clamped force-move step? That could be one part of the issue.

https://github.com/jdoleary/Spellmasons/assets/68413161/627e904e-3a7f-41c5-ad1e-9b10bdcbc63e


_
Units aren't moving as far as they are predicted to move:

https://github.com/jdoleary/Spellmasons/assets/68413161/87968a74-60ee-4d05-a784-92166bad1adb

SoulMuncher:
Explosive arrow prediction is wrong, and shows dealing damage when it doesn't.

I think this is because explosive arrow's explode function is not awaited. I had a similar issue while working on my old PR #183

SoulMuncher:
> can't reproduce: Pushing units through corners.
I can't repro the push through walls either anymore. If I see it again, I'll make sure to create a save

jdoleary:
Note to self: The discrepancy with movePrediction and final location is because the velocity falloff doesn't take the deltaTime into account

jdoleary:
I _think_ i just fixed the incorrect predictions.  I still have to resolve the arrow jumping issue tho

jdoleary:
I'm going to leave the weird arrow wall thing for a separate issue since it isn't directly related to these changes.
@SoulMuncher can you QA check me one last time on this PR and review the code to see if you think I made any errors please? I think it's ready and solves the major problem that was the inspiration of this PR in the first place.

jdoleary:
@SoulMuncher I'm merging this for now so I can release a beta update and test it in the coming week.
When you return please take a look at this PR and help me see if I missed anything or if there are any outstanding issues.  Thanks!

PR #309, opened 2023-12-28
End Turn Sfx plays when an ally ends their turn as well

### LINKED ISSUE #237: Add sound effect when allies end their turn in multiplayer to cue other players
I think this will smooth out the experience of waiting for your allies to end their turn.

SoulMuncher:
Resolves #237

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -2459,7 +2459,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
       return
     }
     // Don't play turn sfx when recording or for auto-ended dead players
-    if ((!globalThis.isHUDHidden && !document.body?.classList.contains('hide-card-holders')) && (player.unit.alive || !player.unit.alive && player.endedTurn)) {
+    if ((!globalThis.isHUDHidden && !document.body?.classList.contains('hide-card-holders')) && (player.unit.alive || (!player.unit.alive && player.endedTurn))) {
       playSFXKey('endTurn');
     }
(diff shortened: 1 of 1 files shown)


src/Underworld.ts
```diff
@@ -2456,13 +2452,16 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
...
     }
-
     if (this.turn_phase != turn_phase.PlayerTurns) {
       // (A player "ending their turn" when it is not their turn
       // can occur when a client disconnects when it is not their turn)
       console.info('Cannot end the turn of a player when it isn\'t currently their turn')
       return
     }
+    // Don't play turn sfx when recording or for auto-ended dead players
+    if ((!globalThis.isHUDHidden && !document.body?.classList.contains('hide-card-holders')) && (player.unit.alive || !player.unit.alive && player.endedTurn)) {
```
jdoleary, line 2462: This boolean logic is faulty
`player.unit.alive || !player.unit.alive && player.endedTurn`
It needs an extra set of parenthesis

jdoleary (changes requested):

src/Underworld.ts
SoulMuncher: > @jdoleary This boolean logic is faulty

The && operator has higher precedence than the || operator, so this is effectively
`... && (player.unit.alive || (!player.unit.alive && player.endedTurn))`
which is the implementation I was going for. I will add the parenthesis for clarity, though.

jdoleary (approved):

Issue #310, opened 2023-12-28
Prevent player units from being cleaned up with Unit.cleanup


jdoleary:
unless explicitly overridden (like during underworld clean up)

Should Unit.cleanup not set the id to -1.

Spells like bone shrapnel and harvest shouldn't have to individually check that they're not cleaning up a player unit.  This should be handled by the cleanup function

jdoleary:
Regarding `  unit.id = -1;` there might not be id conflicts any more.  Look into the cleanup and load functions to make sure

jdoleary:
TODO: syncUnits should:
- NOT remove any existing player's units (even if there are discrepancies)
- Sync the state of player units
- Sync the state of identical non player units
- Remove units not in the serialized units array (so long as they are not player units)
- Create missing units from the serialized array (careful not to overwrite ids of existing units)

jdoleary:
Closed in #313

PR #313, opened 2023-12-29
Safe cleanup

### LINKED ISSUE #306: What happens if you use `bone shrapnel` on an ally player unit
will it break for next level because it cleans them up?

jdoleary:
This PR reworks the sync functions.  Syncing is complicated because sometimes you need to add, remove, or overwrite.
Additionally, we never want to remove certain objects (such as the player unit).

This PR also protects against the player unit from getting cleaned up (for example, in Harvest or BoneShrapnel)

Fixes #306, #310

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ update/
 .vs/
 .vscode/
 public/spellmasons-mods/types
+.DS_Store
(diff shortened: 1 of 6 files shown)


jdoleary:
Automated tests can be run with `npm test Syncronization.test.ts`
@SoulMuncher I'm going to merge this, but can you QA and review it when you return please?

PR #314, opened 2023-12-29
Colorblind support

jdoleary:
In progress system for customizable unit outlines to support colorblind accessibility.
Still needs menu controls


jdoleary:
TODO: Persist changes to localStorage

PR #353, opened 2024-01-07
Impact Damage Fix

### LINKED ISSUE #336: Prediction vs Game Loop impact damage
Impact damage is different in prediction and gameloop

SoulMuncher:
Closes #336 

Force Moves were using deltaPosition (previously true velocity), aka distance moved that frame, to calculate impact damage. This meant that if a prediction used 16ms delta time, and the game loop used 4ms, the game loop would have 1/4th the deltaPosition and thus way less impact damage. They now both use velocity.

The other half of the problem was that even though they're both using velocity now, the velocity itself can be different because of the different simulation time stamps. If a collision would happen at 24ms, the game loop simulates every 10ms, and the prediction loop simulates every 16ms, they are going to register the collision at 30ms and 32ms respectively, so they'll have different velocities at that point. The prediction velocity would have fallen off 2ms longer. I solved this by predicting when the collision would happen in advance, and then estimating the velocity at the time of collision.

estimatedCollisionVelocity has a very very small (<0.1%) margin of error left. I left a comment about it. It's technically possible that there's a 1 damage difference between prediction and game loop, but this is only going to happen once every 1000 high-velocity force moves and 1 damage is very rarely going to be noticeable or the difference between life and death, so I'm going to say it's fine for 1.27.

### TODO - More QA - Specifically Multiplayer

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -485,10 +485,10 @@ export default class Underworld {
       // Check to see if unit has falled out of lava via a forcemove
       Obstacle.tryFallInOutOfLiquid(forceMoveInst.pushedObject, this, prediction);
     } else if (isForceMoveProjectile(forceMoveInst)) {
-      const newPosition = Vec.add(pushedObject, velocity);
+      const newPosition = Vec.add(pushedObject, deltaPosition);
       pushedObject.x = newPosition.x;
       pushedObject.y = newPosition.y;
-      if (math.distance(forceMoveInst.pushedObject, forceMoveInst.startPoint) >= math.distance(forceMoveInst.endPoint, forceMoveInst.startPoint)) {
+      if (math.sqrDistance(forceMoveInst.pushedObject, forceMoveInst.startPoint) >= math.sqrDistance(forceMoveInst.endPoint, forceMoveInst.startPoint)) {
         // Projectile is done if it reaches or goes beyond its end point
         return true;
       }
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -427,36 +427,6 @@ export default class Underworld {
         pushedObject.x = newPosition.x;
         pushedObject.y = newPosition.y;
       }
-      for (let other of aliveUnits) {
-        if (other == forceMoveInst.pushedObject) {
-          // Don't collide with self
-          continue;
-        }
-        // The units' regular radius is for "crowding". It is much smaller than their actual size and it is used
-        // to ensure they can crowd together but not overlap perfect, so here we use a custom radius to detect
-        // forcePush collisions.
-        // Only allow instances that are flagged as able to create second order pushes create new pushes on collision or else you risk infinite
-        // recursion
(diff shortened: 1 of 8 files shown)


src/Underworld.ts
```diff
@@ -394,22 +394,24 @@ export default class Underworld {
       // and is "complete"
       return true;
     }
-    const trueVelocity = Vec.multiply(deltaTime, velocity);
-    const trueMagnitude = Vec.magnitude(trueVelocity);
-    if (trueMagnitude <= 0.1) {
+    // Represents the positional offset for this simulation loop
+    // AKA the Vec2 distance to travel
+    const deltaPosition = Vec.multiply(deltaTime, velocity)
+    if (Vec.sqrMagnitude(deltaPosition) < 0.01) {
```
jdoleary, line 400: Why is sqrMagnitude used here instead of magnitude?

src/jmath/Vec.ts
```diff
@@ -75,8 +75,11 @@ export function dotProduct(p1: Vec2, p2: Vec2): number {
   const angle = clockwiseAngle(getAngleBetweenVec2s(origin, p1), getAngleBetweenVec2s(origin, p2));
   return magnitude(p1) * magnitude(p2) * Math.cos(angle);
 }
+export function sqrMagnitude(p: Vec2): number {
+  return p.y * p.y + p.x * p.x;
+}
```
jdoleary, line 80: I'm confused about this function and it's purpose.  Some comments here would be helpful for me.

src/Underworld.ts
SoulMuncher: Just a little optimization. It has no other differences or impact on the game 👍 See my other comment for more info

src/jmath/Vec.ts
SoulMuncher: sqrMagnitude is magnitude without the sqrt() function. It's primarily used for optimization, as the sqrt() function is very demanding computationally, compared to the basic operators.

If you want to compare the lengths of two vectors, it's faster to compare sqrMagnitudes. If you want to check whether something is within a set distance you're supposed to compare (sqrMagnitude to distance * distance) instead of (magnitude to distance).

Modern computers are good enough that it's not a big deal 99.99% of the time, so usually I don't bother with it. Magnitude is usually a little more readable and nicer to work with. In the case of runForceMove however, sqrMagnitude might be a noticeable optimization, since it could be called 100 times each on 100 different units.

If anything needs the optimization, it's runForceMove 🤣

src/jmath/Vec.ts
```diff
@@ -1,12 +1,15 @@
...
+import { LineSegment } from "./lineSegment";
 import { distance, similarTriangles, lerp } from "./math";
 import { prng, randInt } from "./rand";
 export interface Vec2 {
   x: number;
   y: number;
 }
 
+// TODO - Might fix them later. Not for 1.27.
+
```
jdoleary, line 12: What is "them"?

jdoleary (approved):
Awesome work!  This one is good to go.  Just 2 inline comment related questions and then I'll merge it!

jdoleary (approved):
Great work!

PR #359, opened 2024-01-09
ref: Give hotseat players different client ids

### LINKED ISSUE #269: If you try to load a hotseat game in a multiplayer lobby, you can't ready up

jdoleary:
This is important because sometimes players are looked up by their client ids (like in freeze's add()).

Also this will allow saved hotseat games to be loaded in an only multiplayer lobby.

Fixes #269

### DIFF
--- a/src/HeadlessServer.ts
+++ b/src/HeadlessServer.ts
@@ -79,8 +79,11 @@ class HostApp implements IHostApp {
     // Automatically overridden when passed into pie.startServer
     sendData: (msg: string) => void = () => { };
     overworld: Overworld;
+    soloMode: boolean;
     constructor() {
         this.overworld = makeOverworld(this);
+        // HostApp is run on a server for multiplayer and therefore is never in soloMode.
+        this.soloMode = false;
         new Underworld(this.overworld, this.overworld.pie, Math.random().toString());
     }
     onData(data: any) {
--- a/src/Overworld.ts
+++ b/src/Overworld.ts
@@ -100,7 +100,12 @@ export function ensureAllClientsHaveAssociatedPlayers(overworld: Overworld, clie
             console.log(`Setup: Create a Player instance for ${clientId}`)
             for (let i = 0; i < globalThis.numberOfHotseatPlayers; i++) {
                 const config = globalThis.hotseatPlayerConfig?.[i];
-                const player = Player.create(clientId, underworld);
+                let nextClientId = clientId;
+                // Hotseat players should not share the same client id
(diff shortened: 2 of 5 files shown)


SoulMuncher:
@jdoleary
This PR is an improvement and doesn't seem to introduce any _new_ bugs, but there are still some significant problems, and we may want to implement a larger fix for 1.28. There are many significant differences between how clientIds are handled on hotseat vs in real multiplayer, and it's probably a good idea to minimize these differences (see below). I can and would like to take care of these issues within the refactor I'm working on, since they are somewhat related and currently giving me some trouble. For 1.27, this PR seems good.

Finding player by index or id:
- [Underworld.ts 2538] const playerIndex = numberOfHotseatPlayers > 1 ? this.hotseatCurrentPlayerIndex : this.players.findIndex((p) => p.clientId === clientId);
- [Upgrade.ts 96] const playerUniqueIdentifier = globalThis.numberOfHotseatPlayers > 1 ? player.name : player.clientId;
- [rand.ts 97] const playerUniqueIdentifier = !player ? '0' : (globalThis.numberOfHotseatPlayers > 1 ? player.name : player.clientId);

Network Messages
- These still use the device clientId as "fromClientId" and this results in an incorrect "fromPlayer" for most network messages
- This design forces the above exceptions, and causes broken behavior where similar exceptions are not implemented (as is the case with admin commands like "Super Me", which makes them target the wrong player)

Overworld Structure
- overworld.clients doesn't include other hotseat players. (This is probably a good thing, just noting it here as a potential difference/source of trouble. These cases can be handled manually, I think, and should only come into play when doing setup such as creating player objects in overworld.ts 95 ish)

jdoleary:
Re: your comment above, does this mean that the PR is ready to go given that you're fixing the other stuff in your refactor or should I wait to merge this?

jdoleary:
Also, `overworld.clients` is meant to be specifically a list of wsPie clients so I __think__ it should not include the hotseat players >= 2

SoulMuncher:
> @jdoleary Re: your comment above, does this mean that the PR is ready to go given that you're fixing the other stuff in your refactor or should I wait to merge this?

Yes, I think this PR is good to go. I don't think it introduces any new bugs, and I intend to fix any further issues in my refactor

SoulMuncher:
> Also, `overworld.clients` is meant to be specifically a list of wsPie clients so I __think__ it should not include the hotseat players >= 2

Yeah agreed. I think that makes sense logically

PR #370, opened 2024-01-10
Jan 10 Fixes

### LINKED ISSUE #368: Burst shows VFX at 0 damage

### LINKED ISSUE #372: Target Similar on dead unit doesn't target units from another faction
Two dead golems, one from ally faction, one from enemy faction. Target Similar should target both, but does not.

### LINKED ISSUE #373: Swap should move player to location of last targeted unit
> When I cast [connect - swap] or [cone - swap], I would expect to move to the last unit targeted by the spell, but my player moves to the location of the first targeted unit instead.

Might be good to add a prediction line too?

SoulMuncher:
Closes #368 
Closes #372
Closes #373 - Possible now that targeting order is predictable

- Targeting spells add new targets in a predictable manner (similar to the animation, based on distance rather than the order the units were first created)
- Some code convenience things like sortClosestTo()

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -398,7 +398,7 @@ export default class Underworld {
     // Represents the positional offset for this simulation loop
     // AKA the Vec2 distance to travel
     const deltaPosition = Vec.multiply(deltaTime, velocity)
-    if (Vec.sqrMagnitude(deltaPosition) < 0.01) {
+    if (Vec.magnitude(deltaPosition) < 0.1) {
       // It's close enough, return true to signify complete 
       return true;
     }
@@ -489,7 +489,7 @@ export default class Underworld {
       const newPosition = Vec.add(pushedObject, deltaPosition);
       pushedObject.x = newPosition.x;
       pushedObject.y = newPosition.y;
-      if (math.sqrDistance(forceMoveInst.pushedObject, forceMoveInst.startPoint) >= math.sqrDistance(forceMoveInst.endPoint, forceMoveInst.startPoint)) {
+      if (math.distance(forceMoveInst.pushedObject, forceMoveInst.startPoint) >= math.distance(forceMoveInst.endPoint, forceMoveInst.startPoint)) {
         // Projectile is done if it reaches or goes beyond its end point
         return true;
       }
@@ -3425,7 +3425,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     const withinDistance: HasSpace[] = [];
     const potentialTargets = this.getPotentialTargets(prediction);
     for (let entity of potentialTargets) {
-      if (math.sqrDistance(entity, target) <= distance * distance) {
+      if (math.distance(entity, target) <= distance) {
         withinDistance.push(entity);
       }
     }
@@ -3440,7 +3440,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     const withinDistance: Pickup.IPickup[] = [];
     const pickups = (prediction && this.pickupsPrediction) ? this.pickupsPrediction : this.pickups;
     for (let pickup of pickups) {
-      if (math.sqrDistance(pickup, target) <= distance * distance) {
+      if (math.distance(pickup, target) <= distance) {
         withinDistance.push(pickup);
       }
     }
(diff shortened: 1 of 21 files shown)


src/Underworld.ts
```diff
@@ -3425,7 +3425,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     const withinDistance: HasSpace[] = [];
     const potentialTargets = this.getPotentialTargets(prediction);
     for (let entity of potentialTargets) {
-      if (math.distance(entity, target) <= distance) {
+      if (math.sqrDistance(entity, target) <= distance * distance) {
```
jdoleary, line 3428: I dunno about this.
I'd like to see some stats on this as an efficiency gain.  I'm not sure that it's worth the performance for the tradeoff of being confusing to read.
Can you do some perf testing on distance vs sqrDistance and specifically on
1 sqrDistance and 1 `distance*distance` vs just `distance()` please?

src/entity/units/priest.ts
```diff
@@ -105,9 +105,10 @@ const unit: UnitSource = {
       return [];
     }
     const numberOfAlliesToRez = unit.isMiniboss ? 3 : 1;
-    return resurrectableUnits(unit, underworld).filter(u => {
-      return Unit.inRange(unit, u)
-    }).slice(0, numberOfAlliesToRez);
+    return resurrectableUnits(unit, underworld)
+      .filter(u => Unit.inRange(unit, u))
+      .sort((a, b) => math.sqrDistance(a, unit) - math.sqrDistance(b, unit))
```
jdoleary, line 110: Let's extract this to a util function called `sortClosest` that is meant to be passed into .sort since it's used in many places.  The name will give some added context as to why we're using sqrDistance instead of distance.
Usage:
```
.sort(sortCosestTo(unit))
```
Sample implementation:
```
// returns a function to be passed into a .sort() callback
function sortCosestTo(point: Vec2){
return (a,b) => math.sqrDistance(a, point) - math.sqrDistance(b, point));
}
```
We could do the same with Unit.inRange for convenience:
```
// returns a function to be passed into a .filter() callback
function filterInRangeOf(unit){
   return (target) => math.distance(unit, target) <= unit.attackRange;
}
```

SoulMuncher: Good idea with the sortClosestTo function, will do!

We could do the same for the range one, but it kinda feels like a duplicate of the Unit.inRange() function so I'm not sure we should

src/Underworld.ts
SoulMuncher: @jdoleary 
Wow, I'm glad you asked for this! I did some performance testing and I'm confused / VERY surprised by the results. I ran some tests where I calculated sqrDistance() and distance() a up to a million times each in a large number of trials.

With few loops and trials, the results were kind've sporadic and crazy. performance.now wasn't precise enough to measure the time spent calculating. With higher trial and loop counts, the results get more consistent, but not very promising overall. In the best test, sqrDistance was only about 11% faster. In the worst, it was 3% slower. In most trials... it was 0.5-2 percent faster.

This indicates two things
1) For a distance optimization like this to be noticeable, we'd have to calculate it thousands and thousands of times to notice a performance hit.
2) The optimization that sqrDistance provides is totally negligible

So I was wrong. My understanding had always been that using the squared versions of these functions was 5 or even 10 times faster, but maybe that's not the case with modern tech. JavaScript has some pretty awesome auto-optimization, so maybe that's what's going on here, and if so, there's no reason to make changes that decreases the readability of the codebase. I think you're right that we should just use distance in these places, and I'll be reverting most if not all of these changes.

SoulMuncher: I think I'm going to remove all of the sqrDist and sqrMag stuff I implemented over the last 2ish days. 1% faster calculations aren't worth the maintainability and readability of the code. I'm very disappointed that it ended up being as unhelpful as it was, when I thought I was making a nice improvement, but I'm simultaneously relieved we don't have to deal with the sqrX stuff or worry about the performance impact of using magnitude/distance calculations here and there 😅

jdoleary: No problem! I'm glad that we have some hard data to back up the decision here!

jdoleary (approved):
Looks good!

PR #379, opened 2024-01-13
doc: Concrete Roadmap

jdoleary:
@SoulMuncher Would like to get your thoughts on these.

### DIFF
--- a/docs/old/TasksRoadmap.md
+++ b/docs/old/TasksRoadmap.md
@@ -1,0 +1,71 @@
+As a follow up to https://docs.google.com/document/d/11MwVPXINi27snarC922hPczfOX_OnlthyJfMlIcJ-HE/edit
+This document lays out some concrete tasks and a rough timeline.
+
+## Concrete Tasks
+- New Systems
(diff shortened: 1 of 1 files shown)


SoulMuncher:
Regarding Last Will 2, Poison 2, and Suffocate 2:

I’m not sure every spell needs to have an upgrade, and if it does it should be mechanically different or provide some new interesting behavior in most cases. Direct statistical upgrades for slash and arrows make some sense, because they are basic/core/weak spells at first, but beyond that I feel like unnecessary upgrades makes the game worse. I don’t like the idea of diluting the spell pool with duplicates or encouraging players to use one specific powerful spell instead of being creative with multiple small ones. I think delaying a “quantity” functionality behind an upgrade is also kind’ve lame :P

Other than that, it looks good. I don't think the turn refactor is going to take too much longer so we should be able to fit quite a bit more into 1.28 than planned here, but we’ll adjust as it gets completed. It could take shorter/longer than expected.

jdoleary:
This has all been converted into Issues already so I'm closing this

Issue #394, opened 2024-01-17
Toxic urn visual lags behind effect sometimes


jdoleary:
> Blue — Yesterday at 9:56 PM
> its seems like the visual sometimes doesnt work for the toxic urn but the ability goes off in multiplayer. honestly it might just be lag? it sometimes goes off like many turns after visually 
>

SoulMuncher:
Related? #399 

I was able to reproduce this by spamming a few arrows at some nearby units, and then spamming some arrows at the urns, and then spamming more arrows at other nearby units, as quickly as I could and without waiting for spell resolution.

The urns were deleted from my screen long before they were ever hit by arrows, as if killing the other units had caused a sync that deleted the urns. Maybe the urns were already destroyed on headless because headless simulated all of the force moves in advance, while the real game loop had to wait for each arrow promise one by one?

SoulMuncher:
Should be fixed with #402 and #410 ?

PR #402, opened 2024-01-21
fix: Arrow spell should await all arrows' arrival

### LINKED ISSUE #399: Server Crash: Arrows
https://steamcommunity.com/app/1618380/discussions/0/7091547412719649496/

jdoleary:
Potentially resolves #399, needs more validation

### DIFF
--- a/src/cards/arrow.ts
+++ b/src/cards/arrow.ts
@@ -100,6 +100,7 @@ export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPi
       }
 
     }
+    await underworld.awaitForceMoves();
     return state;
   }
 }


SoulMuncher:
This change may help fix some of the desync that the reporter of #399 mentioned though. They said that some units were magically turning into other units, but I imagine this was just a normal desync issue with some units getting rez'd/killed/moved after sync.

But I think #399 itself is a separate issue: If we look at issue #399 as a network message problem, it has not been resolved. I'm still able to queue up a large number network messages in advance and I can do things like enter the portal on my client without it immediately going to the next level because it's waiting for previous network messages to resolve, and this backup of network messages can cause errors.

I also wonder if this network message issue is related to #394 . See my last comment there for more info

jdoleary:
Okay thanks for the extra context.  I will merge this as it will fix part of the issue but we'll still be on the lookout for the other

Issue #407, opened 2024-01-22
Slash is not Awaited Correctly


SoulMuncher:
(Note: video is sped up by 10% due to issue file size limits)

https://github.com/jdoleary/Spellmasons/assets/68413161/2f263995-253b-445d-9b8a-8018913d71e1

SoulMuncher:
The game will also have to wait for a long timeout if the Slash is refunded (ex. due to targeting a pickup)

jdoleary:
@SoulMuncher This has been fixed, can you please verify?

SoulMuncher:
Re-opening because we forgot to fix heavy and mega slash too lol

PR #410, opened 2024-01-24
Fixed Slash Await

SoulMuncher:
Extention of #407 

Fixes an issue that caused Slash to have an unusually long delay after completion / before other spells in the chain would trigger.

### DIFF
--- a/src/cards/slash.ts
+++ b/src/cards/slash.ts
@@ -32,56 +32,42 @@ const spell: Spell = {
     sfx: 'hurt',
     description: ['spell_slash', damageDone.toString()],
     effect: async (state, card, quantity, underworld, prediction) => {
-      let animationDelaySum = 0;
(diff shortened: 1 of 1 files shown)


jdoleary:
@SoulMuncher 
I added one commit that ensures the animations finish (slash animation was being cleared as soon as the next spell began - this was from a very old line #8ef4a0), can you do another round of QA on this please?

SoulMuncher:
Good catch! Any forseeable issues outside of slash you want me to QA?

jdoleary:
> Good catch! Any forseeable issues outside of slash you want me to QA?

I couldn't think of any other spells that would need to clear containerSpells, so unless you think of any, then no, nothing extra

SoulMuncher:
Looks good! @jdoleary

PR #420, opened 2024-01-29
New Spell: Potion Shatter

### LINKED ISSUE #416: Idea: 'Shatter Potion" - You can damage potions to break them
You can damage potions to break them, which could have some interesting effects:

It could add a special effect to water/liquid, such as healing, additional mana regen, or 0 stamina cost movement
It could apply the potion effect to all units in a small radius
Other ideas?

SoulMuncher:
Closes #416 

.

## TODO
- Art / Thumbnail
- Animation
- SFX

### DIFF
--- a/src/cards/potion_shatter.ts
+++ b/src/cards/potion_shatter.ts
@@ -1,0 +1,46 @@
+import { CardCategory } from '../types/commonTypes';
+import { playDefaultSpellSFX } from './cardUtils';
+import { Spell, refundLastSpell } from './index';
+import { CardRarity, probabilityMap } from '../types/commonTypes';
+import { baseExplosionRadius } from '../effects/explode';
+
+export const potionShatterId = 'Potion Shatter';
+const baseEffectRadius = 50; //baseExplosionRadius / x?
+const spell: Spell = {
+  card: {
+    id: potionShatterId,
+    category: CardCategory.Blessings,
+    supportQuantity: true,
+    manaCost: 15,
+    healthCost: 0,
+    expenseScaling: 1,
+    probability: probabilityMap[CardRarity.COMMON],
+    thumbnail: '',
+    sfx: '',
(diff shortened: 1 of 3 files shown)


src/cards/potion_shatter.ts
```diff
@@ -0,0 +1,59 @@
...
+    manaCost: 15,
+    healthCost: 0,
+    expenseScaling: 1,
+    probability: probabilityMap[CardRarity.SPECIAL],
+    thumbnail: '',
+    sfx: '',
+    description: ['spell_potion_shatter', baseEffectRadius.toString()],
+    effect: async (state, card, quantity, underworld, prediction) => {
+      // .filter: only target potions
+      const targets = state.targetedPickups.filter(p => p.name.includes("Potion"));
```
jdoleary, line 27: I believe this will fail to work with non-english localizations.  I'd use a list of Ids instead

src/entity/Pickup.ts
```diff
@@ -658,12 +658,6 @@ export const pickups: IPickupSource[] = [
             );
           }
         }
-
-        // Now that the player unit's stamina has increased,sync the new
-        // stamina state with the player's predictionUnit so it is properly
-        // refelcted in the stamina bar
-        // (note: this would be auto corrected on the next mouse move anyway)
-        underworld.syncPlayerPredictionUnitOnly();
```
jdoleary, line 666: The stamina bar still updates correctly without this?

src/cards/potion_shatter.ts
SoulMuncher: Right! Forgot to update that. Thanks!

jdoleary (approved):

src/entity/Pickup.ts
SoulMuncher: In my testing yes I think so. I couldn't make it not work.

PR #423, opened 2024-02-01
Mana Vampire Changes

### LINKED ISSUE #350: deathmason boss stops doing things if bitten by mana-vampire.
> SneakyMinion
> OP
>  — Today at 7:20 AM
> he just stops moving/casting spells or doing anything the moment he gets bitten even just once.
> 
> so despite still having 140-180 mana, seems to just give up on life (to be fair, I give up the moment a mana-vampire bites me too; instant rage-quit)

SoulMuncher:
Closes #350 

Mana Vampire now steals mana instead of mana max, and uses that mana to heal itself each turn.

Also modularized restoration effects for healing and mana.

Tested in Singleplayer and w/ loading

## TODO
- Update Mana Vampire description to fit new behavior

### DIFF
--- a/src/cards/add_heal.ts
+++ b/src/cards/add_heal.ts
@@ -5,13 +5,14 @@ import { CardCategory } from '../types/commonTypes';
 import { playDefaultSpellSFX } from './cardUtils';
 import { Spell } from './index';
 import { CardRarity, probabilityMap } from '../types/commonTypes';
+import { healUnits } from '../effects/heal';
 
-export const heal_id = 'heal';
+export const healCardId = 'heal';
 const healAmount = 30;
 
 const spell: Spell = {
   card: {
-    id: heal_id,
+    id: healCardId,
     category: CardCategory.Blessings,
     sfx: 'heal',
     supportQuantity: true,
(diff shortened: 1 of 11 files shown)


SoulMuncher:
@jdoleary There's a lot of places we use the potionPickup effect for mana and instead of calling addOneOffAnimation, we have to do this whole process to change the color of the effect to blue, and this difference is not only messier but also causes problems such as a longer animation delay due to not having the keyframe ending or something.

Do you have a quick solution to this?

jdoleary:
Can you elaborate on the longer animation delay problem?
As for the messiness, we could always extract it to a function.

This is the block that you're referring to, yes?
  ```
  const animationSprite = addPixiSpriteAnimated('spell-effects/potionPickup', unit.image.sprite, {
                loop: false,
                onComplete: () => {
                  if (animationSprite?.parent) {
                    animationSprite.parent.removeChild(animationSprite);
                  }
                }
              });
              if (animationSprite) {

                if (!animationSprite.filters) {
                  animationSprite.filters = [];
                }
                // Change the health color to blue
                animationSprite.filters.push(
                  new MultiColorReplaceFilter(
                    [
                      [0xff0000, manaBlue],
                    ],
                    0.1
                  )
                );
              }
```

SoulMuncher:
It looks like these codeblocks are actually outdated and there's a clean way of doing it with the regular function now. Going to clean this up where applicable.

jdoleary (approved):

PR #460, opened 2024-02-13
Waves Fix

### LINKED ISSUE #451: Stepping in a portal won't advance the game if there are enemies still alive
I know this combination should never happen in a live game but this could maybe cause a softlock if it ever somehow happened.  I found this while trying to skip a level in admin testing.  If all players are in portals, the game should advance no matter if any enemies or allies are left alive.

### LINKED ISSUE #452: You should get renewed mana when a next wave spawns, but your mana remains the same.
> Isneverthere
> OP
>  — Today at 2:25 AM
> Yeah. I switched servers and it worked. The second wave started and the only option was to end turn with no mana and we were insta blasted. Does the new wave just spawn on last kill?
https://discord.com/channels/1032294536640200766/1206122453617741874/1206138867028201545

- [ ] Test with hotseat

Reproduced in both multiplayer and singleplayer

SoulMuncher:
Closes #452

Separates handleLevelProgress() into smaller/more dedicated functions and moves wave spawning to endTurnFullCycle() alongside boss spawning. Reasoning: New waves shouldn't interrupt the progression of the game state like portal spawning or game over logic does, and should only really occur between turns.

Also improves logging and added a couple helper functions to get remaining units and players

Things to note:
- Wave counter now starts at 1 instead of 0
- Some console logs have been changed from .log to .debug, so they are only visible when needed / when verbose logging is selected. Broader progression logs are still shown as normal

QA - Singleplayer, Multiplayer, Loading, Waves, Deathmason level, combinations of on death/freeze/etc. killing last enemy and player at the same time

## TODO
Done

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -152,7 +152,7 @@ export default class Underworld {
   // The index of the level the players are on
   levelIndex: number = -1;
   isTutorialRun: boolean = false;
-  wave: number = 0;
+  wave: number = 1;
   // for serializing random: prng
   RNGState?: SeedrandomState;
   turn_phase: turn_phase = turn_phase.Stalled;
@@ -1907,7 +1907,7 @@ export default class Underworld {
     // Reset kill switch
     this.allyNPCAttemptWinKillSwitch = 0;
     // Reset wave count
-    this.wave = 0;
+    this.wave = 1;
     // Reset has spawned boss boolean
     this.hasSpawnedBoss = false;
   }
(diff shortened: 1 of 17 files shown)


src/Underworld.ts
```diff
@@ -2125,6 +2125,22 @@ export default class Underworld {
...
     return { x, y };
   }
+  getPlayerFactions() {
+    // Returns all players' current faction
+    return this.players.filter(p => p.clientConnected).map(p => p.unit.faction);
+  }
+  getRemainingPlayerUnits(): Unit.IUnit[] {
+    // Returns all living units currently controlled by a player
+    return this.players.filter(p => p.isSpawned).map(p => p.unit).filter(u => u.alive);
+  }
```
jdoleary, line 2135: What about `isConnected`?

src/Underworld.ts
```diff
@@ -2125,6 +2125,22 @@ export default class Underworld {
     );
     return { x, y };
   }
+  getPlayerFactions() {
+    // Returns all players' current faction
+    return this.players.filter(p => p.clientConnected).map(p => p.unit.faction);
```
jdoleary, line 2130: I'd recommend deduplicating this array:
https://stackoverflow.com/a/9229821/4418836

so that it'd only ever be ['ally','3rd faction', 'etc'] rather than ['ally', 'ally', 'ally', '3rd faction', 'etc', 'etc']

jdoleary (approved):
Great work,
Note, the Deathmason brothers still don't spawn (used to be that 3 deathmasons spawned after the first was killed) but that is a separate bug and this PR looks great so I'll merge it.

PR #467, opened 2024-02-13
Deathmason On Death Event Fix

### LINKED ISSUE #449: First deathmason dying no longer kills remaining enemies
It also no longer spawns his brothers.

SoulMuncher:
Closes #449

Ensures bosses spawned with introduceBoss use the normal spawnEnemy function and have originalLife = true
Moves the Deathmason onDeathEvent from the level gen logic to the introduceBoss logic, since it's no longer spawned during level gen.

QA: Singleplayer, Loading

Note: If a player loads a save in which the Deathmason has already spawned without the onDeathEvent, it will not be added.

## TODO
Done

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -4226,6 +4226,7 @@ async function introduceBoss(unit: UnitSource, underworld: Underworld) {
     unit.unitProps,
     underworld
   );
+  newBossUnitInstance.originalLife = true;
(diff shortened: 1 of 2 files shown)


jdoleary (approved):
Looks great! I don't mind that the onDeathEvent isn't backwards compatible, I don't see that as an issue 👍

PR #487, opened 2024-02-14
test: Reports live in console

jdoleary:
if promises are left unresolved by the time
the game moves onto casting the next card.effect.

This is a work in progress and could stand to be cleaned up, but it should be super useful for finding cards with dangling unawaited promises.

Note, some promises are intentionally unawaited like heal's floatingText.  The effect of the spell need not await the completion of the floating text, so it is ignored.


SoulMuncher:
Did a quick test in singleplayer/headless with a few different spells and intentionally forgotten awaits, looks good!

Two things:
- We probably shouldn't do await testing during predictions right? Only for the actual cast and for headless?
- It might be best to use .debug for the test logs and only show the warnings/errors in the console by default

jdoleary:
I addressed your comments and merged this so now we should get console errors if a particular spell has an async issue

PR #491, opened 2024-02-16
Shatter and Radius Improvements

### LINKED ISSUE #456: Bloat added to Plus Radius Bloat
If you apply bloat with several stacks of plus radius, and then later apply bloat to the same unit, it should retain the additional radius

### LINKED ISSUE #476: New Spell: Shatter
A damage spell that deals lots of damage if the enemy is frozen. If Shatter kills the enemy, it unleashes ice shards in all directions

Requires 'Freeze'

SoulMuncher:
Closes #476 
- Implemented a version of Shatter: Removes freeze from target unit to deal 40 damage to it and all surrounding units. Damage increased by quantity. Explosion radius increased by freeze stacks.

Closes #456 
- Plus Radius will now additively stack on separate applications of bloat.

Extra:
- Also improves `state.aggregator.radius` handling, allowing "Plus Radius" to have a variety of effects on different spells.
  - Instead of aggregating a flat radius increase, the aggregated radius boost is now more like a "counter", counting how many times the radius has been boosted, and this number can be used to situationally increase the radius of other spells (additively, multiplicatively, with percentages, stacked linearly with quantity or other variables, etc.)
  - Instead of Plus Radius adding+50 radius to all spells equally, it now adds one radius boost, which is interpreted differently by each spell. This ensures Plus Radius is useful with all "area" spells, instead of being mandatory on low range spells (potion shatter) and near useless on long range spells (target column)
  - This also allows for linear stacking with quantity where preferred, I.E. The radius of Shatter or Potion shatter will increase by some amount per quantity or plus radius, with each granting the same radius increase, making it more predictable

## TODO
- i18 Description?

### DIFF
--- a/src/cards/contaminate.ts
+++ b/src/cards/contaminate.ts
@@ -49,8 +49,10 @@ interface CurseData {
 
 // separate function to handle synchronous recursion and animation - avoids long wait times
 async function contaminate(casterPlayer: IPlayer | undefined, unit: IUnit, quantity: number, radiusBoost: number, underworld: Underworld, prediction: boolean) {
-  const adjustedRadiusBoost = quantity - 1 + radiusBoost;
-  const adjustedRange = baseRange * (1 + (0.5 * adjustedRadiusBoost)) * (casterPlayer?.mageType == 'Witch' ? 1.5 : 1);
+  // Contaminate does not get extra radius from quantity,
+  // but can instead chain an additional time
+  // +25% range per radius boost
+  const adjustedRange = baseRange * (1 + (0.25 * radiusBoost)) * (casterPlayer?.mageType == 'Witch' ? 1.5 : 1);
 
   // the units to spread contaminate from
   // initally just this unit, then only the latest additions to the chain
--- a/src/cards/contaminate.ts
+++ b/src/cards/contaminate.ts
@@ -14,9 +14,10 @@ import { IPlayer } from '../entity/Player';
 import { Modifier } from './util';
 import { summoningSicknessId } from '../modifierSummoningSickness';
 import { corpseDecayId } from '../modifierCorpseDecay';
+import { baseExplosionRadius } from '../effects/explode';
 
 export const contaminate_id = 'contaminate';
-
+const baseRange = baseExplosionRadius;
 const spell: Spell = {
   card: {
     id: contaminate_id,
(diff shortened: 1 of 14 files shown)


src/cards/contaminate.ts
```diff
@@ -47,8 +48,9 @@ interface CurseData {
 }
 
 // separate function to handle synchronous recursion and animation - avoids long wait times
-async function contaminate(casterPlayer: IPlayer | undefined, unit: IUnit, underworld: Underworld, extraRadius: number, prediction: boolean, quantity: number) {
-  const range = (COLLISION_MESH_RADIUS * 4 + extraRadius) * (casterPlayer?.mageType == 'Witch' ? 1.5 : 1);
+async function contaminate(casterPlayer: IPlayer | undefined, unit: IUnit, quantity: number, radiusBoost: number, underworld: Underworld, prediction: boolean) {
+  const adjustedRadiusBoost = quantity - 1 + radiusBoost;
```
jdoleary, line 52: why "- 1"?

jdoleary (commented):
Looks good overall.
How was it tested?  Was each of the modified files retested?

src/cards/contaminate.ts
SoulMuncher: The common formula I'm using is
adjustedRange = baseRange * (1 + (x% * radiusBoost)

or essentially
adjustedRange = baseRange + (some% of baseRange per **_additional_** quantity/plus radius)

The range of a spell doesn't usually double for every quantity/plus radius, so we count the initial quantity and the "extra" quantities separate, hence the - 1. The first quantity will give the entire base range, and is handled in the "baseRange * 1" part of the equation, while later quantities and "plus radius" only give a portion of that range, and are handled in the "+ (0.x * radiusBoost) part of the equation, where radiusBoost usually = (_extra_ quantity + state.aggregator.radiusBoost)

SoulMuncher: All this being said, Contaminate's radius isn't supposed to scale with quantity anyways, so I fixed that
(Contaminate is a special case where extra quantity causes chaining instead of increased aoe radius: Increased radius will come from "Plus Radius" and the Witch class only.)

jdoleary:
Looks great! I'd say let's try with thinner particles on the explosion so it looks different from the ice urn's explosion (more like shards of ice rather than a wave of frost)

SoulMuncher:
> Looks great! I'd say let's try with thinner particles on the explosion so it looks different from the ice urn's explosion (more like shards of ice rather than a wave of frost)

I removed the bone shrapnel and default explode vfx from shatter and gave it its own vfx explosion instead. Let me know what you think/how it can be improved.

jdoleary (approved):

PR #493, opened 2024-02-16
Meteor

### LINKED ISSUE #473: New Spell: Meteor
Meteor falls on top of enemies doing splash damage around them

SoulMuncher:
Closes #473 

Summons a meteor from the sky which deals massive damage in an area on impact

## TODO
- VFX Improvements
- i18 description

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -3583,7 +3583,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
       targetedPickups: [],
       castLocation,
       aggregator: {
-        radius: 0,
+        radiusBoost: 0,
       },
       initialTargetedUnitId,
       initialTargetedPickupId
--- a/src/cards/arrow.ts
+++ b/src/cards/arrow.ts
@@ -85,6 +85,7 @@ export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPi
             pushedObject,
             startPoint: casterPositionAtTimeOfCast,
             endPoint: endPoint,
+            speed: 1.5,
             doesPierce,
             ignoreUnitIds: [state.casterUnit.id],
             collideFnKey
--- a/src/cards/arrow_fork.ts
+++ b/src/cards/arrow_fork.ts
@@ -37,7 +37,7 @@ const spell: Spell = {
         for (let newAngle of [Math.PI / 12, -Math.PI / 12, 2 * Math.PI / 12, -2 * Math.PI / 12, 3 * Math.PI / 12, -3 * Math.PI / 12]) {
           const angle = getAngleBetweenVec2s(projectile.startPoint, unit) + newAngle;
           const castLocation = getEndpointOfMagnitudeAlongVector(unit, angle, 10_000);
-          arrowEffect(1, arrowCardId)({ cardIds: [regularArrow.card.id], shouldRefundLastSpell: false, casterPositionAtTimeOfCast: unit, targetedUnits: [], targetedPickups: [], casterUnit: unit, castLocation, aggregator: { radius: 0 }, initialTargetedPickupId: undefined, initialTargetedUnitId: undefined }, regularArrow.card, 1, underworld, prediction, false);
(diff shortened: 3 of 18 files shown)


jdoleary:
We should use particles to animate a meteor coming down from the sky, I can show you how to use the particle editor (https://pixijs.io/pixi-particles-editor/#) and how you import it's settings into the game
(Note the size of the particles shows as larger(smaller?) in the editor than they do in game when you export and include the particle settings.

jdoleary:
In order to animate the particles trailing off the meteor, you should be able to reuse projectile functions (in arrow) but without the collision) and attach the particle emitter to the projectile (like how they're attached to the deathmason).
I'd say have the meteor come in from an angle above (like a 30 degree angle down)

jdoleary:
I LOVE it!  Very fun to call down meteors.
Some notes:
- The particles should be "left behind" as the meteor travels through space - not streaming out of it.  See how this is done with the deathmason particles (if you push the deathmason, you'll see it).
- Have the meteor start up higher and move faster
- I want multiple meteors from a single case to have a delay (like multiple arrows) rather than coming down all at the same time)
- use public/images/empty.png for the image

SoulMuncher:
@jdoleary 

> * I want multiple meteors from a single case to have a delay (like multiple arrows) rather than coming down all at the same time)

This makes the spell very unpredictable due to the AoE knockback, and it'll make the in-game predictions inaccurate unless we do a lot of special work with it like we did with arrows (even then I'm not 100% sure it will work/feel good). Lots of things to consider here with awaiting, overlapping force pushes, and hitting/not hitting initially targeted/not targeted things.

> * use public/images/empty.png for the image

"public/images/empty.png" doesn't work with Image.create since it starts in the other image/ folder. I didn't see images/empty used anywhere else in the codebase so I'm not really sure how to do that

jdoleary:
Right, I see what you're saying.  Okay for now, ignore what I said about empty.png and separating the timings of the meteors.
A few notes.  Because they start at 0,0 there's some particles that always end up at 0,0.
I'd like the meteors to start up much higher.  They appear and come in so fast that you can't really enjoy the anticipation of the collision and it looks weird if you're zoomed way out.

SoulMuncher:
> some particles that always end up at 0,0

Having some trouble fixing this one. It's happening due to Pixi's innate emitter-position-interpolation. For some reason the particle emitters always start interpolating from 0,0, which is why they're appearing there even though the meteors are initialized elsewhere, and emitters attached directly to them. It happens with the Deathmason as well: when he spawns/teleports you'll see interpolated particles between 0,0 and his spawn, or between point A and B of the TP.

**There is a function on the emitter called resetPositionTracking() which is supposed to fix this**, but I haven't had much luck with it. This was the fix I just attempted in the meteor script:

```
if (wrapped) {
      const particleFollower = {
        displayObject: wrapped.container,
        emitter: wrapped.emitter,
        target,
      }
      particleFollower.emitter.resetPositionTracking();
      underworld.particleFollowers.push(particleFollower);
    } else {
      console.warn('Failed to create meteor particle emitter');
    }
```

Will keep looking for a solution.

Meteors now spawn 5x higher and with almost 3x the total travel time (Making them move faster but also giving more anticipation. Looks much better when zoomed out!)

SoulMuncher:
Also, would you have any idea how to keep the emitter particles around after the meteor has landed? I think that would do a TON for the vfx. Currently they're being destroyed along with the force move projectile, since the emitter no longer has a target, but the already emit particles should persist.

I tried detaching the particles and setting a new target/disabling emission via an OnProjectileCollision event with the meteors but I coudn't quite figure it out

jdoleary:
I'm going to make a separate issue for these two remaining so we can move on and I'll investigate those issues further later.
For now let's move on so we don't get stuck on this.

https://github.com/jdoleary/Spellmasons/issues/523

With that in mind, do you feel that meteor is ready to merge.  I'm pretty happy with it

SoulMuncher:
Yes, other than those 2 particle behaviors I think it's ready!

jdoleary (approved):
I'm very happy with this! Great work!

PR #521, opened 2024-02-23
Target Curse

### LINKED ISSUE #481: New Spell: Target Curse
A curse that makes the next cast spell also target the cursed unit

SoulMuncher:
Closes #481 

Target Curse is a spell/modifier that adds the cursed unit to the targets list of the next cast spell. Quantity increases the number of spells it affects.

## TODO
- i18n description


src/cards/target_curse.ts
```diff
@@ -0,0 +1,148 @@
...
+        }
+      }
+
+      const potentialTargets = underworld.getPotentialUnitTargets(prediction);
+      // Add all other target-cursed enemies to targets
+      for (const unit of potentialTargets) {
+        if (!targets.includes(unit) && unit.modifiers[targetCurseId]) {
+          addTarget(unit, state);
+        }
+      }
```
jdoleary, line 48: There needs to be something in Underworld.ts that adds TargetCursed units to _any_ next spell, not just spells preceeded with target curse.
The idea is that I could cast target curse on a unit near me and then your next spell would also affect the unit I cursed.
This can be done around the same place that the logic for "Protected" is run (Underworld.ts line 3669)
This will be hard coded in Underworld.ts for this spell in particular.  If this pattern becomes shared among more spells we can refactor it then, but for now I'm confortable with it being hard coded.

src/cards/target_curse.ts
```diff
@@ -0,0 +1,148 @@
...
+const spell: Spell = {
+  card: {
+    id: targetCurseId,
+    category: CardCategory.Targeting,
+    supportQuantity: true,
+    manaCost: 40,
+    healthCost: 0,
+    expenseScaling: 1,
+    probability: probabilityMap[CardRarity.RARE],
+    thumbnail: 'spellIconTargetCurse.png',
```
jdoleary, line 28: Missing the `d` at the end

src/cards/target_curse.ts
```diff
@@ -0,0 +1,148 @@
...
+
+      return state;
+    },
+  },
+  modifiers: {
+    add,
+  },
+};
+
+export async function animateTargetCurse(newTargets: Vec2[]) {
```
jdoleary, line 63: OMG this animation is AWESOME.  Well done

src/cards/target_curse.ts
SoulMuncher: I don't totally understand this design:
- The only use case I can come up with for singleplayer is a more complicated version of conserve? [Mass target + Target Curse] => turn => [Target Cursed + Damage] => turn => [Target Cursed + Damage] => ... Maybe I could mass target + x quantity of target curse and then several dmg/heal spells, but why wouldn't I just put that dmg/heal after the initial targeting combo? It would only be useful as a between-turn mana buffer, unless I'm missing something
- Anti-Multiplayer: There's few scenarios where one player casting target curse is going to ruin the next player's spell. I don't forsee a global targeting effect feeling good with more than 2 players, and even with two you need to play much more slowly and with much more communication, to avoid healing enemies, damaging allies, triggering bloat in the wrong place, pushing players to their death, etc. Might also create an exploit similar to Timemason-SendMana where the strat is for one player to [Mass Target + Target Cursed] every level, and another player to deal damage based off of that free targeting (free mana, essentially).
- Seems a little more unintuitive and effect is too automatic. A lot of players may not understand how or why the spell is behaving this way because they're not controlling it or interacting with it. It's also not really a persistent "curse", so the name feels misleading

I think a lot of people were hoping for a more persistent effect/lasting curse, and such is the core identity of the spell. It's for cheap persistent targeting when you want to hit the marked units, instead of being for adding a bunch of additional targets up front.

SoulMuncher: Hmm, maybe [this](https://discord.com/channels/1032294536640200766/1032294537235812404/1181632622187774112)..? I can't find any examples of people describing exactly how they want it to behave, it just seemed like the assumed behavior and use case was something more persistent / more like a curse

> I'll target him for you and you hit him next turn

My implementation doesn't allow that kind of game sequence, so I see how it falls short in that area 🤔 I'll go ahead and move the targeting logic to CastCards so it affects the next spell, rather than requiring Target Curse, and we'll see how it goes.

Also 'Forbidden' rarity seems fitting, but maybe we should release spells at a lower rarity initially, that way people see and get a chance to try the new stuff? I've played so many games since we've made Target Similar Forbidden and I've not gotten it once... Lol.

SoulMuncher: Ready for review again 👍

SoulMuncher: @jdoleary Ready for review again 👍

PR #528, opened 2024-02-28
Target Injured

### LINKED ISSUE #482: New Spell: Target Injured
A spell that adds a unit with less than maximum health to the targeted list.

SoulMuncher:
Closes #482 

- Targets the next 2 * Quantity units with the most missing HP. If multiple units have the same missing HP, it will prioritize the closest one. Ignores units at full hp.

## TODO
- i18n description

### DIFF
--- a/src/cards/target_injured.ts
+++ b/src/cards/target_injured.ts
@@ -33,7 +33,7 @@ const spell: Spell = {
       const addedTargets = [];
 
       for (const target of potentialTargets) {
-        if (target.health <= adjustedHealthThreshold && target.unitSubType != UnitSubType.DOODAD) {
+        if (target.alive && target.health <= adjustedHealthThreshold && target.unitSubType != UnitSubType.DOODAD) {
           addedTargets.push(target);
         }
       }
--- a/src/cards/target_injured.ts
+++ b/src/cards/target_injured.ts
@@ -1,16 +1,16 @@
 import { addTarget, EffectState, ICard, Spell } from './index';
 import { CardCategory, UnitSubType } from '../types/commonTypes';
-import { Vec2, lerpVec2 } from '../jmath/Vec';
+import { Vec2 } from '../jmath/Vec';
 import * as Unit from '../entity/Unit';
 import { CardRarity, probabilityMap } from '../types/commonTypes';
-import { easeOutCubic } from '../jmath/Easing';
 import * as config from '../config';
 import * as colors from '../graphics/ui/colors';
 import Underworld from '../Underworld';
 import { calculateGameDifficulty } from '../Difficulty';
+import { distance } from '../jmath/math';
 
 export const targetInjuredId = 'Target Injured';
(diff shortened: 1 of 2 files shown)


jdoleary:
You could just use the green circle to show their targeted

src/cards/target_injured.ts
```diff
@@ -0,0 +1,142 @@
...
+    healthCost: 0,
+    expenseScaling: 1,
+    probability: probabilityMap[CardRarity.RARE],
+    thumbnail: 'spellIconTargetInjured.png',
+    requiresFollowingCard: true,
+    description: 'spell_target_injured',
+    allowNonUnitTarget: true,
+    effect: async (state: EffectState, card: ICard, quantity: number, underworld: Underworld, prediction: boolean, outOfRange?: boolean) => {
+      // 20, 40, 80, 160, 320...
+      const adjustedHealthThreshold = healthThreshold * Math.pow(2, quantity - 1) * calculateGameDifficulty(underworld);
```
jdoleary, line 29: What's the need for adjustedHealthThreshold?  Sorting them by missing health should be sufficient

SoulMuncher: Maybe I'm misunderstanding.

This spell targets all units under a certain health threshold and adds them to the targets list in order of missing health.

Are you wanting it to instead target the next X units with the lowest hp or something?

The adjusted health threshold is just the threshold after factoring in quantity and health multipliers due to difficulty scaling (target injured would be useless in myltiplayer without this)

jdoleary: Target Injured should target units in order of "most missing health"

200/300 (100 missing)
5/90 (85 missing)
120/140 (20 missing)
55/60 (5 missing)

jdoleary: How would this be useless in multiplayer?

SoulMuncher: > @jdoleary How would this be useless in multiplayer?

You can ignore that point: The spell idea was originally written/discussed as targeting things under a certain health threshold, and in that implementation multiplayer/difficulty scaling becomes a big deal, requiring multiple extra casts of the spell to have the same effect.

Health scaling isn't relevant with the implementation you're describing now, which targets the next X units with the most missing health instead of using thresholds

Sorry for the confusion 🙂 Will update the spell behavior

SoulMuncher: New Behaviour:
- Targets the next 2 * Quantity units with the most missing HP. If multiple units have the same missing HP, it will prioritize the closest one. Ignores units at full hp.

jdoleary: ah okay, I didn't remember the old description.  Yeah i think this works better :)

Is this branch ready for review?

jdoleary (approved):
It feels cheap for targeting 2 whole units.  What do you think about just 1 unit per quantity?

Either way, it looks great.  Approved.  Merge when ready

PR #535, opened 2024-03-02
Recall

### LINKED ISSUE #475: New Spell: Recall
A movement spell that creates a mark on the ground. If the mark already exists, you teleport to that mark

SoulMuncher:
Closes #475 

Also fixes an issue where players can cast 'Teleport' out of bounds.

## TODO
- VFX (Change portal to something else?)
- i18n


jdoleary:
Use `images/spell-effects/recall` for the vfx.

jdoleary:
I should've been more specific: not for the thumbnail, the thumbnail was fine as it was.  Instead of the portal.  Create a new pickup that uses `spell-effects/recall` instead of the portal image.  It should be a pickup so it can be moved around but it should have no pickup effect

SoulMuncher:
Oh gotcha! Thats on me. Will revert last commit and change the pickup image

jdoleary:
Looks great! The animation was funky so I fixed a few frames.

PR #541, opened 2024-03-05
Soul Shard

### LINKED ISSUE #478: New Spell: Soul Shard
The caster splits a piece of their soul into the target. When the target takes damage so does the caster, but if the caster dies, they will reincarnate in the body of the shard-haver

This should be do-able with modifiers

SoulMuncher:
Closes #478 

The caster splits a piece of their soul into the target. All damage that would be dealt to the shard bearer is instead redirected to the shard owner (I.E. the caster), but if the shard owner dies, they will reincarnate in the body of the nearest shard-bearer, and all of their soul shards will be consumed.

## TODO
- Damage Redirection FX slightly broken
- Maybe multicolor replace filter on teleport FX
- #570 
- i18n


jdoleary:
A few thoughts:
- I think Soul Shard should only be castable on allies.  Otherwise, you can essentially instakill any enemy you shard when you yourself die
- When the caster dies, let's have a particle (like when summoners teleport) flies to the one they will reincarnate in and when it arrives, they die and the caster reincarnates

SoulMuncher:
> Let's make Soul Shard only be castable on allies. Otherwise, you can essentially instakill any enemy you shard when you yourself die
I really like the idea of the spell being castable on enemies too. It leaves room for more creative use and emergent gameplay. At the same time though, maybe you're right. I'm not sure how we'd balance it without significantly redesigning the spell. 

> When the caster dies, let's have a particle (like when summoners teleport) flies to the one they will reincarnate in and when it arrives, they die and the caster reincarnates.
Sounds good!

> When you die, only one (not all) of the sharded NPCs should die. The one you reincarnate in.
It should already do this. I can't repro any other behavior

> When you reincarnate, maybe you should keep the mana you had when you die? I'm unsure on this one
Yeah maybe, I'm not really sure either. I lean towards no, and also having the player revive at 1 hp?

@jdoleary
Random idea combining points 1 and 4: Another implementation of the spell would have it redirect the damage you take to the sharded entities, and then when the sharded entity dies it triggers the teleport/res/heal/mana restoration. May not fit the use case you have in mind? Maybe we turn these into two separate spells, Soul Shard and Soul Shield, and tweak each accordingly to be offensive/defensive. I actually really like that...

jdoleary:
Like you mentioned in Discord, if we allow you to cast soul shard on enemies you totally become unkillable.  If you cast it on an ally, you will potentially run out of allies and lose the level.
Does this not solve that problem?
Also I agree, 1 hp and no mana, it's too powerful to not have a consequence.
The offensive idea is cool but isn't that essentially Soul Bind?

> I can't repro any other behavior
Okay retested: Soul Shard shows prediction death for all sharded NPCs if you self kill, but will correctly only kill one.
- [ ] Prediction should only show that the one that you kill dies.

I added a little commit that prevents the game over screen from coming up just by setting `unit = alive` at the top of the soul shard onDeath event (I abandoned awaiting onDeath cause it was super messy and I fixed the main issue elsewhere)

- [ ] Ran into a bug: the onDeath isn't persisted though a save.  If you soul shard a unit, save and then load and then die.  You just die (not reincarnate) and the soul sharded npc is invincible cause they redirect all damage to your dead character.

SoulMuncher:
> Okay retested: Soul Shard shows prediction death for all sharded NPCs if you self kill, but will correctly only kill one.
 Prediction should only show that the one that you kill dies.
 
Okay, I did some more extensive testing and I'm seeing this now. It only happens when you're targeting and damaging a bunch of enemies. Might have to do with prediction awaiting the on death effect differently and resolving before the next unit.takeDamage. I'll check it out 👍

jdoleary:
New idea for the VFX:
When soul shard is cast, have a particle (color of the player's robes) travel from the caster to the targets as if it's pieces of the player's soul.
When the shard-haver dies, I notice their corpse is kind of covering the player, so let's clean up the corpse and just leave behind a huge blood splatter as if you reincarnated out of their body.  You should be able to customize startBloodParticleSplatter something like: `startBloodParticleSplatter(underworld, targetVec2, unit, { maxRotationOffset: Math.PI * 2, numberOfParticles: 300 });`

Also some balancing: since soul shard is so powerful, maybe it should cost some max health.  Like 10% every cast?  Thoughts?

SoulMuncher:
> New idea for the VFX: When soul shard is cast, have a particle (color of the player's robes) travel from the caster to the targets as if it's pieces of the player's soul.

Sounds very cool! Might be nice if it made the unit played their damage animation as well? This idea reminds me of the [2nd phase transition](https://youtu.be/IBAKjviBElo?si=Vxz-R2LMicMO7ebu&t=93) for the final boss in Diablo 3

.

> Also some balancing: since soul shard is so powerful, maybe it should cost some max health. Like 10% every cast? Thoughts?

Hm, I'm not sure. Max-stat penalties feel bad. People have commented on this with Death Wager and Mana Vampires. I don't think it's too op with the 1hp/0 mana respawn, high mana cost/scaling, and only targeting allies.

SoulMuncher:
> @jdoleary The offensive idea is cool but isn't that essentially Soul Bind?

In the same way that Soul Shard is essentially Soul Bind, I guess.

My idea was:
Soul Shard: Redirects damage you take to them. If they die, you resurrect in their place.
Soul Shield: Redirects damage they take to you. If they die, they resurrect in your place.

They can be further distinguished by incentivizing offensive/defensive behaviors. Ex.
- Soul Shard gives you full hp/mp/stamina when the sharded entity dies
- Soul Shield gives you a full turn of invulnerability

Not totally convinced on the idea myself though, just brainstorming something that sounded cool thematically 😄

SoulMuncher:
Did some testing with Soul Shard. It feels preeeetty bad. The onDeath behavior is very unreliable, and instantly resurrecting still has its own weird edge cases. Ex. Dying to ancients, dying to mid-flight ghost arrows, dying to bloat (sometimes), stuff like that. The worst part is when something unexpected happens, instead of being a "woah that was weird" moment, you just straight up lose the game.

Possible solution: Just resurrect the player at the start of their next turn. I can easily socket this modifier into the game-end logic and have it continue the game loop if they would be revived. This is how I imagine any kind of "undying" modifier would work anyway. I feel like this would be much cleaner and also allow for a neat resurrect animation.

I'll go ahead and start on it! @jdoleary

jdoleary:
Hmm good thought, I'm curious to see how it turns out!

SoulMuncher:
Okay, lots of changes in, let me know what you think.
I'm might do a little more testing and tweaking for balance.

jdoleary:
Just spent some time playing around with this and it is AWESOME!  Such a cool spell.  I think we may have to have some kind of effect to show that the damage is being transferred though cause I guarantee players will use it and then be confused why they're taking damage.  (The shard moving animations are awesome btw).  So let's do something different for the damage transfer: how about we use the lightning effect from Bolt but change the color to the same dark red you've been using? (and make it last less long, like 200ms)

src/entity/Unit.ts
```diff
@@ -177,6 +180,7 @@ export function create(
       staminaMax,
       attackRange: config.UNIT_BASE_RANGE,
       isMiniboss: false,
+      isPrediction: false,
```
jdoleary, line 183: undefined is falsy so this could remain an optional property and would function identically (except it wouldn't be serialized which would save some bandwidth).  Why was this made explicit?

src/modifierSoulShardOwner.ts
```diff
@@ -0,0 +1,88 @@
...
+            const trailColorEnd = colors.convertToHashColor(colors.healthBrightRed);
+            // Trails from bearers
+            let promises = [];
+            for (let shardBearer of allShardBearers) {
+              promises.push(makeManaTrail(shardBearer, unit, underworld, trailColorStart, trailColorEnd));
+            }
+            await Promise.all(promises);
+
+            // Trail from shard owner
+            await new Promise<void>(resolve => oneOffImage(unit, 'units/summonerMagic', containerUnits, resolve));
```
jdoleary, line 48: Let's tint this or use MultiColorReplaceFilter to get it to be the same color as the particle

src/cards/soul_shard.ts
```diff
@@ -0,0 +1,193 @@
...
+          const graphics = globalThis.predictionGraphics;
+          if (graphics) {
+            const lineColor = colors.healthDarkRed;
+            for (let target of targets) {
+              graphics.lineStyle(4, lineColor, 1);
+              // Temp VFX - Draw X
+              graphics.moveTo(target.x + 10, target.y + 10);
+              graphics.lineTo(target.x - 10, target.y - 10);
+              graphics.moveTo(target.x - 10, target.y + 10);
+              graphics.lineTo(target.x + 10, target.y - 10);
```
jdoleary, line 45: This is cool, how about just the shape of a diamond instead of an "X" and we'll call it good?

SoulMuncher: Sounds good!

src/entity/Unit.ts
SoulMuncher: When I tried to use unit.isPrediction in line 143 of soul_shard.ts it gave me an error that it might be undefined:

```ts
function remove(unit: Unit.IUnit, underworld: Underworld) {
  const soulShardModifier = unit.modifiers[soulShardId];
  if (soulShardModifier) {
    removeShardOwner(soulShardModifier.shardOwnerId, underworld, unit.isPrediction); // <--- HERE
  }
}
```

Wait, could I have just used the !! symbol here?🤦‍♂️I'll revert this......

jdoleary:
Note to self: Left TODO:
Multicolor replace filter for the teleport fx and the bolt fx

SoulMuncher:
TODO: Verify that Soul Shard works with contaminate. It may not spread the shard owner id, which could cause problems.

#570

jdoleary (approved):

PR #543, opened 2024-03-08
On Deal Damage Event

### LINKED ISSUE #470: New Spells: Empower / Enfeeble
A blessing/curse that makes a unit deal more/less damage

SoulMuncher:
closes #470

Empower/Enfeeble: Increase/Decrease a unit's damage stat by 5 per stack.

## TODO
- FX / Descriptions
- Ensure intended behavior with synergies like split, clone, and purify
  - Split is bugged, which is a different issue. This seems to function well enough in other situations.


src/effects/heal.ts
```diff
@@ -24,25 +24,33 @@ export async function healUnits(units: Unit.IUnit[], amount: number, underworld:
...
 
-export async function healUnit(unit: Unit.IUnit, amount: number, underworld: Underworld, prediction: boolean, state?: EffectState, useFx: boolean = true) {
+export async function healUnit(source: any | undefined, unit: Unit.IUnit, amount: number, underworld: Underworld, prediction: boolean, state?: EffectState, useFx: boolean = true) {
   const units = [unit];
-  return await healUnits(units, amount, underworld, prediction, state, useFx);
+  return await healUnits(source, units, amount, underworld, prediction, state, useFx);
 }
 
-export async function healManaUnits(units: Unit.IUnit[], amount: number, underworld: Underworld, prediction: boolean, state?: EffectState, useFx: boolean = true) {
+export async function healManaUnits(source: any | undefined, units: Unit.IUnit[], amount: number, underworld: Underworld, prediction: boolean, state?: EffectState, useFx: boolean = true) {
```
jdoleary, line 53: source unused?

src/jmath/moveWithCollision.ts
```diff
@@ -30,6 +30,7 @@ export function isForceMoveUnitOrPickup(x: ForceMove): x is ForceMoveUnitOrPicku
 }
 export type ForceMoveProjectile = ForceMove & {
   type: ForceMoveType.PROJECTILE;
+  source: any;
```
jdoleary, line 33: let's type these to something specific, not `any`

src/jmath/moveWithCollision.ts
```diff
@@ -42,6 +43,7 @@ export function isForceMoveProjectile(x: ForceMove): x is ForceMoveProjectile {
 
 interface ForceMoveProjectileArgs {
   pushedObject: HasSpace;
+  source: any;
```
jdoleary, line 46: let's type these to something specific, not `any`

jdoleary: It can be an union type if needed
https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-types

jdoleary (commented):
Looking good, a few comments

SoulMuncher:
@jdoleary The initial idea was that anything can be a source of damage. Pickups pass themselves as the source when applying their effect, for example. There may be times when we want to pass a modifier or something else as a source as well, which would have been helpful for soul bind's implementation, since it needs to know when it's the thing doing damage. It might be nice to keep generic for future rune effects as well.

Do you foresee any specific issues with leaving it as "any"?

jdoleary:
I'd rather be more specific.  I'm not a fan to setting things up that "might be useful in the future".  I'd rather just put them in when they are needed because then we'll have a better idea of the actual requirements.  `any` is dangerous because it breaks typesafety and can introduce runtime-bugs.

src/cards/enfeeble.ts
```diff
@@ -0,0 +1,64 @@
...
+
+function add(unit: Unit.IUnit, underworld: Underworld, prediction: boolean, quantity: number = 1) {
+  const modifier = getOrInitModifier(unit, enfeebleId, { isCurse: true, quantity, }, () => {
+    //no first time setup
+  });
+
+  unit.damage += statChange * quantity;
+
+  // Ensure damage doesn't go below 0
+  unit.damage = Math.max(unit.damage, 0);
```
jdoleary, line 51: It'd be funny if making a unit deal negative damage meant they healed their target.  I know this doesn't make any sense, but let's try it anyway to pleasantly surprise players who try it

jdoleary: We could even add the heal animation when damage is dealt that it negative outside of the heal spell to show that it's on purpose.

jdoleary (commented):
This one feels almost ready!

src/cards/bloat.ts
```diff
@@ -92,6 +92,7 @@ const spell: Spell = {
       const radiusBoost = unit.modifiers[id]?.radiusBoost;
       const adjustedRadius = getAdjustedRadius(radiusBoost);
       explode(unit, adjustedRadius, damage * quantity, getAdjustedPushDist(radiusBoost),
+        unit.modifiers[id]?.sourceUnit, // TODO - Implement source unit for modifiers
```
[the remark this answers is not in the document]
SoulMuncher: I would say no. The player unit (or bloat caster) is the one that inflicted the damage here, so should be the one whose onDealDamage events should trigger. Practical examples: The player unit has 50% critical strike chance, or a blessing that makes all of their outgoing damage inflict poison. The player should apply that crit/poison through the bloat, because it is _their_ spell/outgoing damage. The bloated unit is not the one dealing damage, and should not have their own effects trigger.

I found that storing a unit in a modifier leads to all kinds of problems, and doing the id workaround like I did with soul shard is somewhat complicated, so I left it like this for now. The Poison and Bloat modifiers do not actually reference their source unit currently.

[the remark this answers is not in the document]
jdoleary: What's complicated about storing the unit id and fetching the unit object later when it's needed?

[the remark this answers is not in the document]
SoulMuncher: "Messy" is probably a better word. I ran into some problems with prediction/realtime id's being different than one another, so if you're not careful you can run into some really weird behavior. You don't always have access to the prediction variable either so in some cases it can be tricky to access it (which is why you saw me accessing the unit.isPrediction variable in the Soul Shard PR).

I can definitely do it, it's just a little bit of a workaround and requires a few extra helper functions. I just need to move some of these helper functions to Underworld to make sure it looks clean 👍

src/cards/enfeeble.ts
SoulMuncher: > We could even add the heal animation when damage is dealt that it negative

Just to clarify: Do you want the heal animation built into Unit.takeDamage()?

jdoleary: I think so... If damage dealt is ever negative we __always__ play the heal animation right?
I'm thinking of potions... all heal spells... I think that's it. 
Is there ever any reason why we wouldn't want this?
Maybe it should go in a modifier instead? I'm open to ideas

src/cards/bloat.ts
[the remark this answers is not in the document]
jdoleary: You can use `predictionUnit.real` to backreference the real unit.
So to find a prediction unit that matches a real unit you can do `predictionUnits.find(pu => pu.real.id == searchID)`.
It's a little expensive of a computation cause it potentially has to loop though all units but that's how it can be done.

Alternatively we could add a reference  to the prediction unit from the real unit.  Like, when copyForPredictionUnit is called, we assign the property `u.predictionUnit = newlyMadePredictionUnit` so the real unit can keep a reference to the current prediction unit.
This may actually be the best solution since I'm going to be adding a sync change that will probably make use of this.
(Note: if you do this you'll have to update the unit serialization function to omit the `predictionUnit` property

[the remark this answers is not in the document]
jdoleary: I'll probably finish the sync update today.  So this new `predictionUnit` property would be available to you tomorrow.  I'd say wait for my change and then this should be an easy update for you

src/cards/enfeeble.ts
SoulMuncher: Nope, I love the idea. Just double checking you meant in takeDamage and not in the modifier or something 😁

src/cards/bloat.ts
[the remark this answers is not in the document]
SoulMuncher: Sounds good!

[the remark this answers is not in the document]
jdoleary: Here it is: https://github.com/jdoleary/Spellmasons/pull/577

SoulMuncher:
After putting the heal FX in takeDamage() I noticed the health potion has two overlapping healing sounds fx, since the potion isn't _just_ the glass bottle pickup noise, but also has its own heal effect sound. It's not too noticeable but thought I'd mention it in case you miss it in testing.

SoulMuncher:
@jdoleary 
Empower/Enfeeble don't work on poisoner and players as one might expect. How do you want to handle this currently? If you're not sold on #578 I could exclude them from the target list.

jdoleary:
Without doing #578 yet, can we update poisoners to give as many stacks of poison as their damage (and set their base damage to 1)?

SoulMuncher:
> Without doing #578 yet, can we update poisoners to give as many stacks of poison as their damage (and set their base damage to 1)?

Sure. I'm assuming they're still untargetable by empower and enfeeble?

SoulMuncher:
> Well that would make poisoners targetable with empower and enfeeble. Like if you cast it once they go from 1 stack of poison to 2. Only spellmasons would be untargetable

Okay, I'll make a special case for poisoners in empower/enfeeble. I left other casters untargetable since they wont be affected by damage changes, and I left the spellmasons targetable since your damage stat affects cloned spellmasons.

SoulMuncher:
Should be ready 👍

src/Underworld.ts
```diff
@@ -3558,6 +3558,10 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     const closest = sortedByProximityToCoords[0]
     return closest;
   }
+  getUnitById(id: number, prediction: boolean): Unit.IUnit | undefined {
+    const units = prediction ? this.unitsPrediction : this.units;
+    return units.find(u => u.id == id);
+  }
```
jdoleary, line 3564: I foresee potential issues with this function given that prediction units have inconsistent ids.  Has this been tested with clone / split prediction units?

src/entity/units/poisoner.ts
```diff
@@ -55,11 +56,17 @@ const unit: UnitSource = {
         ).then(async () => {
           // Add projectile hit animation
           Image.addOneOffAnimation(chosenUnit, 'projectile/poisonerProjectileHit');
+          const cardsIds = [];
+          // Casts one stack of poison per damage
+          for (let i = 0; i < unit.damage; i++) {
+            cardsIds.push(poison.poisonCardId);
+            console.log(cardsIds);
```
jdoleary, line 63: remove console.log

jdoleary (commented):
Looking good, 2 comments

src/Underworld.ts
SoulMuncher: Leaving this here per our discord convo:

> I'd just say test it thoroughly (like clone and then bloat and make sure it passes the correct sourceUnit) and if it works fine, then we can merge it
(...)
I see a potential problem as a prediction unit getting an id, the id gets saved in the modifier, and then the prediction unit gets cleaned up or something
(...)
"use .real instead of assuming that prediction units id matched their real counterparts id"
Maybe just leave a comment to that effect with getUnitById

jdoleary (approved):

PR #551, opened 2024-03-17
Fix Targeting Unspawned Players

### LINKED ISSUE #550: bug: you can use target similar on self to kill unspawned players
this induces a game over desync
https://discord.com/channels/1032294536640200766/1070397640765874327/1218310717778694155

SoulMuncher:
closes #550 

Fixes a bug that allowed unspawned players to be targeted, which could cause a desync in the game over state. Adds further support for additional universal targeting catches by passing underworld as a parameter to the addTarget() function

QA Complete

### DIFF
--- a/src/cards/target_similar.ts
+++ b/src/cards/target_similar.ts
@@ -71,7 +71,7 @@ export function targetSimilarEffect(numberOfTargets: number) {
         animators.push({ pos: target, newTargets: newTargets });
       }
       for (let newTarget of newTargets) {
-        addTarget(newTarget, state);
+        addTarget(newTarget, state, underworld);
       }
     }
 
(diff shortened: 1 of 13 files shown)


src/cards/target_similar.ts
```diff
@@ -50,6 +50,11 @@ export function targetSimilarEffect(numberOfTargets: number) {
         // @ts-ignore Find similar units by unitSourceId, find similar pickups by name
         .filter(t => {
           if (isUnit(target) && isUnit(t) && t.unitSourceId == target.unitSourceId) {
+            if (underworld.players.filter(p => !p.isSpawned).map(p => p.unit.id).includes(t.id)) {
+              // Do not allow targeting unspawned players
+              console.log("Filtered out unspawned player from target similar's potential targets");
+              return false;
+            }
```
jdoleary, line 57: Isn't this redundant, now that addTarget omits unspawned players?

jdoleary (commented):
Looks good, one comment

src/cards/target_similar.ts
SoulMuncher: Yes, it's redundant. I left it in this spot as well since unspawned players should probably *never* be targeted, and it'll serve as protection against future spells. Not absolutely necessary though, it comes at the cost of passing in underworld.

jdoleary (approved):

Issue #554, opened 2024-03-17
Level Generates multiple times / Too many upgrade points


jdoleary:
https://discord.com/channels/1032294536640200766/1032294537235812404/1218996298552836236

SoulMuncher:
It could be due to the portal triggering multiple times. I might be able to replicate it by adding an intentional delay in level gen code, so the player is over the portal longer/for multiple frames. It could be that it only happens if that code reaches the thread limit and takes multiple frames to generate, which might happen on slower computers.

SoulMuncher:
I was able to repro this on the current Steam version by casting a few spells before going through the portal. 

Ideas: Might work better with high delay spells like arrows, bleed, and debilitate. Might require at least 1 spell queued up and not currently simulating.

jdoleary:
JordanO — Today at 9:21 PM
Do you have any leads on that?
SoulMuncher — Today at 9:22 PM
That was that one where entering the next level while multiple spells were queued would make it generate extra times
other than that I'm not sure.
JordanO — Today at 9:22 PM
ah right
SoulMuncher — Today at 9:23 PM
But having network messages queued was definitely the cause, spamming debilitates and bleeds was a pretty reliable way to repro it
JordanO — Today at 9:23 PM
That's a good clue
SoulMuncher — Today at 9:23 PM
you can make a save right before going to the next level, after everything has died
JordanO — Today at 9:23 PM
ah nice
that'll be helpful for me to repro

PR #560, opened 2024-03-20
Fix Level Regen

SoulMuncher:
WIP #554 

Update: I was able to repro this by casting a few spells before going through the portal. Might work better with high delay spells like arrows, bleed, and debilitate. Require at least 1 spell queued up and not currently simulating. The repro is consistent under these conditions.

## TODO
- Identify source of bug
  - createLevelSyncronous() being called multiple times?
  - Does failed level generation queue/resolve multiple "create level" network messages
  - Is this because the last player entering the portal can trigger it several times, queueing up several network messages?

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -1932,6 +1932,7 @@ export default class Underworld {
     this.hasSpawnedBoss = false;
   }
   postSetupLevel() {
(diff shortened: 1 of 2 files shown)


SoulMuncher:
Part of #691

PR #569, opened 2024-03-22
Special Handling for Freeze Subsprite

### LINKED ISSUE #568: When the whole party is frozen in the same turn game endlessly skips turns
Waterbending Squirrel
OP
 — Today at 6:09 PM
To replicate bring whole party in one place and use something like Target Circle + Freeze
The game is stuck in endless loop when it says its your turn but you can't move or do any other action, and enemies just permamently attack you.

Note: after being frozen the players that got frozen can't move until someone? does and action. We will test this further and make another bug.

SoulMuncher:
closes #568 

Worth noting: This does not fix the issue of Freeze vfx persisting for a short time after being purified or removed by other means. This only fixes the endless turn looping

## TODO
- Done I think

### DIFF
--- a/src/cards/freeze.ts
+++ b/src/cards/freeze.ts
@@ -53,18 +53,19 @@ const spell: Spell = {
   modifiers: {
     add,
     remove,
-    subsprite: {
-      imageName,
-      alpha: 1.0,
-      anchor: {
-        x: 0.5,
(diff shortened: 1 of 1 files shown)


jdoleary (approved):
Well done!  Great fix :)

PR #576, opened 2024-03-23
wip: initial foray into sending final state WITH spell

jdoleary:


jdoleary:
Note to self: I think what I want to do here is, rather than sending a diff, send SYNC_SOME_STATE payload in the **same message* as the spell from the player that cast.  That way they are guaranteed that the effect of their spell prediction will be what happens when the spell finishes and since it syncs during the SPELL message it shouldn't get interrupted by anything else.

PR #583, opened 2024-03-28
New Spell: Polymorph

### LINKED ISSUE #474: New Spell: Polymorph
Changes a unit into another random unit

Since I love for Spellmasons to suprise people, if players are chaotic enough to self-cast polymorph I think they should be allowed to use the abilities of the unit that they poly into. Maybe even give them hidden stat upgrades (like upgrading damage). This would be a very rewarding suprise. It would also have to disable or remove their regular spells (until they purify to unpolymorph)

Basically the idea is to pleasantly suprise. If you so choose to turn yourself or your ally spellmason into an archer, I'd expect you to then only have arrow spells and be able to play a "hidden game" of Spellmasons as an Archer instead of a wizard. If you poly into a golem, you'll only have a "bite" attack. Yes, super limiting but therein lies the pleasant suprise. It's the natural consequence of polymorphing yourself and I think players will love it.

Also it being reversable via purify will make the mistake more hilarious and forgivable in multiplayer (provided you have it)

### LINKED ISSUE #630: Address todo in polymorph.ts
Address todo in polymorph.ts
      // TODO - Improve probability curve and move to unit test
      // https://github.com/jdoleary/Spellmasons/pull/583#discussion_r1551439805

SoulMuncher:
Closes #474 
Closes #630 

Implements Polymorph, a spell that turns one thing into a different thing.

## TODO
- QA (?)
- Description


SoulMuncher:
Initial design thoughts: Polymorph should be treated as though the enemy has newly spawned as the new unit type.
- Stats should reset fully
- Added modifiers should be retained (shield, bloat, poison, etc.)
- New "Built in" modifiers should be added (I.E. If a Priest morphs into a Vampire, it should gain blood curse, as blood curse is built into the vampire's identity)
- Old "Built In" modifiers should be removed (I.E. If a Vampire morphs into a Priest, it should lose blood curse, as blood curse was an innate quality of the Vampire)

If it works this way, it needs to be well distinguished. If it's not easily distinguishable, the other option is to simply persist all or none of the present modifiers.

jdoleary:
This is looking great so far! Good job catching some of those edge cases like `isModActive`

SoulMuncher:
> @jdoleary 
For polymorph, I don't want it to get too complicated just to decrease the possibility of weird bugs, so err on the side of simple even if it means leaving some stuff out. With that said:
> - Modifiers carry over (don't worry about special exceptions)
> - Purifiable
> - Bosses switch between other bosses (Goru and Deathmason)
> - Spellmasons (player) should change images and stats but still be able to cast
> - Spellmason NPCs should fully change like any other npc
> - Pickups could change to another pickup, similar to alchemize, but to any pickup (like traps to potions or vice versa - blue and red portals too) 
> - Cooler if polymorph were completely random instead of cycling like the potion spell does. Let's do this: Higher chance of polymorphing into something of the same unit budget and lower chance of polymorphing into farther budgets (bellcurve)

src/cards/polymorph.ts
```diff
@@ -0,0 +1,165 @@
...
+              possibleUnitTypes = possibleUnitTypes
+                .filter(u => u.unitProps.isBoss);
+            } else {
+              // Get all units with budget and spawn chance
+              possibleUnitTypes = possibleUnitTypes
+                .filter(u => u.spawnParams && u.spawnParams.probability > 0 && u.spawnParams.budgetCost);
+            }
+
+            // We have to seed this to prevent multiplayer desync
+            const seed = seedrandom(`${underworld.turn_number} -${target.id} `);
```
jdoleary, line 55: add underworld.seed to the front of the seed string

src/cards/polymorph.ts
```diff
@@ -0,0 +1,165 @@
...
+            const chosenIndex = chooseObjectWithProbability(possibleUnitTypes.map((p, index) => {
+              // Units are weighted by their difference in budget.
+              // Units twice as far away in the budget are half as likely to be chosen.
+              const budgetDiff = Math.abs((unitSource?.spawnParams?.budgetCost || 0) - (p.spawnParams?.budgetCost || 0));
+              return { index, probability: (budgetDiff == 0) ? 1 : 1 / budgetDiff }
```
jdoleary, line 60: This is a perfect candidate for unit testing.
Not the actual random picking but the resulting probabilities.

Extract to a function that takes 2 budgets and returns a probability, then test for a variety of possibilities (same, small difference, far difference, negative, 0, etc).

src/cards/polymorph.ts
```diff
@@ -0,0 +1,165 @@
...
+            }), seed)?.index;
+
+            const chosenUnitType = (chosenIndex != undefined) ? possibleUnitTypes[chosenIndex] : undefined;
```
jdoleary, line 63: No need to use .index on line 61 and then use the index on line 63.  Just have chooseObjectWithProbability return `p` and as unit {unit:p, probability: etc} instead of {index, probability} and then do .unit on line 61

src/cards/polymorph.ts
```diff
@@ -0,0 +1,165 @@
...
+          }
+        }
+      }
+
+      return state;
+    },
+  },
+};
+
+function polymorphUnit(fromUnit: Unit.IUnit, toUnitId: string, underworld: Underworld, prediction: boolean): Unit.IUnit | undefined {
```
jdoleary, line 132: We have to handle player units here. If fromUnit is of type player_controlled then we just change the image and don't do Unit.create

src/cards/polymorph.ts
```diff
@@ -0,0 +1,165 @@
...
+function polymorphPickup(fromPickup: IPickup, toPickupSource: Pickup.IPickupSource, underworld: Underworld, prediction: boolean): IPickup | undefined {
+  const pickup = Pickup.create({ pos: fromPickup, pickupSource: toPickupSource, logSource: 'spawnPickup' }, underworld, prediction);
+  if (!prediction) {
+    setTimeout(() => {
+      playSFXKey('spawnPotion');
+      floatingText({ coords: fromPickup, text: toPickupSource.name });
+    }, 500);
+  }
+  return pickup;
+}
```
jdoleary, line 163: Seems odd that the cleanup of the original pickup doesn't happen in this function.  Though I get it because you need to remove it from targets, but those two behaviors definitely belong together (imagine if this function were called in a different context).  I'd say inline the contents of this function for that reason.  Same for the above polymorphUnit.  I don't see any reason that they should be extracted as individual functions, particularly because they don't handle all the logic inherent in polymorphing.

src/entity/Unit.ts
```diff
@@ -184,6 +185,7 @@ export function create(
       staminaMax,
       attackRange: config.UNIT_BASE_RANGE,
       isMiniboss: false,
+      isBoss: false,
```
jdoleary, line 188: I'd rather just check for the unit id if we need to determine "boss-ness" rather than adding a property to every unit.  Remember every property that we add to units needs to be send over the network very frequently for every unit unless it's optional.

src/jmath/rand.ts
```diff
@@ -87,8 +87,8 @@ export function chooseObjectWithProbability<T extends objectWithProbability>(
     (maxProbability, current) => current.probability + maxProbability,
     0,
   );
-  // Choose random integer within the sum of all the probabilities
-  const roll = randInt(1, maxProbability, seedRandomInstance);
+  // Choose random number within the sum of all the probabilities
+  const roll = randFloat(0, maxProbability, seedRandomInstance);
```
jdoleary, line 91: This is a significant change.
Why float instead of int?
With 0 as the lower bound this would take 1-2 (a 50/50 chance) and change it to 0-2 (33/33/33).
With a change like this I'd expect to see an explanation in the PR comments and added tests to cover the new functionality.
My intuition is to keep this as it was, but I'm curious about your reasoning here.

SoulMuncher: Rand float will give the same result as rand int currently does, but will also handle non-integer probabilities if given

For the example you gave, the chance is still 50/50. Two things with a probability of 1 would generate a number between 0 (inclusive) and 2 (exclusive). Anything less than 1 = result of 1, anything less than 2 = result of 2. Both have an equal number of possibilities (0-0.999 and 1-1.999)

I did a little more research to double check before responding, and I also found that [Math.Random() has an exclusive maximum](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), so I may update the surrounding code to reflect that as well (mostly comments and variable names, change <= to < on line 57, and remove the probability != 0 exception I added on line 54)

other source: [Will JavaScript random function ever return a 0 or 1](https://stackoverflow.com/questions/11730263/will-javascript-random-function-ever-return-a-0-or-1)

src/jmath/rand.ts
```diff
@@ -87,9 +70,18 @@ export function chooseObjectWithProbability<T extends objectWithProbability>(
...
+  // Iterate each object and check if the roll is within the probability range
+  for (let x of source) {
+    if (roll < x.probability) {
+      return x;
+    }
+    roll -= x.probability;
+  }
+
+  console.warn("ChooseObjectWithProbability returned nothing. This probably shouldn't happen?: ", roll, source)
+  return undefined;
```
jdoleary, line 84: Good catch on finding that randFloat maxInclusive should actually be maxExclusive - I missed that.

That rename change can remain, however, I want you to leave _chooseObjectWithProbability and chooseOneOfSeeded as they were.  This is long-standing, well-trusted engine code (which also has tests already).  I want to avoid engine code changes unless there is something actually wrong.  I'd rather adapt the code that uses the engine code to meet the limitations of the engine code rather than change the engine code to meet every changing need.  I see that you find it useful to have this work with floats so that polymorph can use `probability: (budgetDiff == 0) ? 1 : 1 / budgetDiff`; however it will be much safer to just modify polymorph slightly so that the probability is an integer such as `Math.floor(1000 / budgetDiff)`.
Especially this late in the development of the game, it is a high-risk decision to change engine code like this.  It might be fine, but it might introduce hidden errors and would require a lot of testing and scrutiny.  Opt for the simpler, safer solution instead (unless of course a bug or problem is found in the engine code).

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+import seedrandom from 'seedrandom';
+import { chooseObjectWithProbability, chooseOneOf, chooseOneOfSeeded, getUniqueSeedString } from '../jmath/rand';
+
+export const polymorphId = 'Polymorph';
+const spell: Spell = {
+  card: {
+    id: polymorphId,
+    category: CardCategory.Curses,
+    supportQuantity: true,
+    manaCost: 40,
```
jdoleary, line 23: This should probably be more expensive, maybe 60 or 80.  What do you think?

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+    healthCost: 0,
+    expenseScaling: 2,
+    probability: probabilityMap[CardRarity.RARE],
+    thumbnail: 'spellIconPolymorph.png',
+    sfx: 'purify',
```
jdoleary, line 28: Oooh I like the idea of using Purify sfx and vfx.  Let's use the purify VFX too but change the color so it looks distinct.  Just use the MultiColorReplaceFilter

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+        refundLastSpell(state, prediction, 'No valid targets. Cost refunded.');
+        return state;
+      }
+
+      playDefaultSpellSFX(card, prediction);
+
+      for (let i = 0; i < quantity; i++) {
+        for (const target of targets) {
+          if (Unit.isUnit(target)) {
+            const unit = target as Unit.IUnit;
```
jdoleary, line 42: This typecast `as Unit.IUnit` _should_ be unnecessary due to the [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates) in the line just above it

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+            // Replace current unit with new unit
+            const newUnit = polymorphUnit(unit, underworld, prediction);
+            if (newUnit) {
+              // Visual aid only
+              if (prediction) {
+                // TODO - Show new unit?
```
jdoleary, line 48: no need to show the new unit, it should be a surprise.  That adds to the fun!

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+          if (modifier?.add) {
+            modifier.add(unit, underworld, prediction, modifierInstance.quantity, modifierInstance);
+          }
+        } else {
+          console.error("Modifier doesn't exist? This shouldn't happen.");
+        }
+      }
+
+      // Cleanup old unit and remove it from targets
+      Unit.cleanup(fromUnit, false);
```
jdoleary, line 141: This line doesn't remove the unit from targets, that is handled in the calling function right now.  Let's pass the effectState into this function and handle the cleanup and adding the new unit to the target array here

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+
+  console.log(toPickupSource);
+  const pickup = Pickup.create({ pos: fromPickup, pickupSource: toPickupSource, logSource: 'spawnPickup' }, underworld, prediction);
+  if (pickup != undefined) {
+    if (!prediction) {
+      playSFXKey('spawnPotion');
+      floatingText({ coords: fromPickup, text: toPickupSource.name });
+    }
+    // Cleanup old pickup and remove it from targets
+    Pickup.removePickup(fromPickup, underworld, prediction);
```
jdoleary, line 189: This line doesn't remove the pickup from targets, that is handled in the calling function right now. Let's pass the effectState into this function and handle the cleanup and adding the new pickup to the target array here

src/cards/polymorph.ts
```diff
@@ -0,0 +1,194 @@
...
+    // ID is required here to ensure that every polymorph changes the seed to prevent looping
+    const seed = seedrandom(`${getUniqueSeedString(underworld)} - ${fromPickup.id}`);
+    toPickupSource = chooseOneOfSeeded(possiblePickupTypes, seed);
+    if (toPickupSource == undefined) {
+      console.error("Polymorph failed to choose a new pickup type.");
+      return undefined;
+    }
+  }
+
+  console.log(toPickupSource);
```
jdoleary, line 181: remove log.

Tip: When i'm testing stuff while I develop with logs I usually put a unique identifier at the front or end such as 
`console.log('jtest', toPickupSource)`.  I do this because
1. It's easy to find them all later and remove the test ones with a simple search
2. If one gets left in accidentally it's easy to find and remove.

I've worked on many projects where a testing log accidentally gets committed and then it's really hard to find because it's logging just an object and so is rather unsearchable in the codebase.

jdoleary (commented):
Looking good!  A few comments

src/cards/polymorph.ts
SoulMuncher: Agreed! This was from when the result was deterministic. Will remove 👍

src/cards/polymorph.ts
SoulMuncher: Personal opinion? Unpredictable effects feel inherently weak to me in Spellmasons. Why would I risk making a level more difficult when I can consistently guarantee a win using more controlled/surefire tactics? Feels like the only use case right now is deleting summoners and min-maxing capture soul once a level is under control.

That being said, I haven't used the spell much in real games, and it is quite good against summoners, so I bumped it up to 60.

jdoleary:
Left todo:
> the multicolorfilter for spell effects

SoulMuncher:
TODO: Quantity broke

jdoleary:
Self-targeting polymorph seems to only ever turn the player into a golem

jdoleary (approved):

PR #590, opened 2024-03-31
New Spell: Merge

### LINKED ISSUE #472: New Spell: Merge
Merges two units together, summing their stats
This will be tricky, how will it interact if used on a Spellmason? Maybe it only impacts current health / mana instead of max?

SoulMuncher:
closes #472

Merges all like-types among currently targeted things. Golems with golems, archers with archers, etc.

## TODO
- QA (?)
- FX (?)


jdoleary:
Merges should stick within factions lines

jdoleary:
FX: Try using particles and the stretch effect that clone uses

jdoleary:
I've already done the translation, just make the description `spell_merge`

src/cards/merge.ts
```diff
@@ -0,0 +1,194 @@
...
+  },
+};
+
+export function mergeUnit(target: Unit.IUnit, unitsToMerge: Unit.IUnit[], underworld: Underworld, prediction: boolean) {
+  // TODO - I forsee a bug that causes modifier addition to change game state before merge is complete
+  // I.E. Target gets a few stacks of suffocate and dies, bloat triggers, bunch of stuff changes.
+  // Loop keeps going and reaches some undefined value, or tries to add modifiers to dead target, etc.
+  for (const unit of unitsToMerge) {
+    // Never merge a player controlled unit
+    if (unit.unitType == UnitType.PLAYER_CONTROLLED) continue;
```
jdoleary, line 82: I feel like _something_ should happen if you try to merge players.  Better to do something cool even if it ends the run than do nothing (fizzle spells are not fun in a game about spell combining).  If you try to merge with Non player spellmasons (like clones) maybe it just adds to your current health/mana/stam rather than max?

If you try to merge with another player... I'm not sure, we can't destroy other players units but I know people are going to try it so it should do something.  Maybe it just steals their current health and mana down to 1 and gives them to the caster?  We should definitely do something to pleasantly surprise players who try it

src/cards/merge.ts
```diff
@@ -0,0 +1,194 @@
...
+}
+
+
+export function mergePickup(target: Pickup.IPickup, pickupsToMerge: Pickup.IPickup[], underworld: Underworld, prediction: boolean) {
+  console.log("TODO - Merge Pickups");
+  for (const pickup of pickupsToMerge) {
+    // TODO - Pickup Merging
+    // Will require pickup rewrite
+
+    // Combine pickup effects / strength?
```
jdoleary, line 139: Let's make this refactor a part of this spell.
Give pickups a "power" property.  Everything currently starts at "1".  Merged potions get bigger and have their power property summed.  This _should_ be a relatively small refactor I think.

jdoleary: Potion effects should be multiplied by their power level just like how 'quantity' works for spells.

src/cards/merge.ts
SoulMuncher: @jdoleary 
Okay, hear me out on this one: When a player would be merged into another unit, it instead spawns a little tiny Spellmason 🤣

jdoleary: Hahaha I love it! Like between them? or how would it work?

src/entity/Pickup.ts
```diff
@@ -58,6 +58,8 @@ export type IPickup = HasSpace & {
   // Pickups optionally have a "time limit" and will disappear after this many turns
   turnsLeftToGrab?: number;
   text?: PIXI.Text;
+  // Boosts the effect of the pickup, default 1
+  power: number;
```
jdoleary, line 62: The copy for each pickup now needs to be changed so that it accounts for the power level.  For example merged health potions still say "heals 50" when it heals 100.

This will mean pickup description will have to be changed to a function that returns a `Localizable` array instead of just the array and the invocation in PlanningView line 907 will have to change from `i18n(globalThis.selectedPickup.description)` to `i18n(globalThis.selectedPickup.description(globalThis.selectedPickup)`.
That way the pickup instance will be available in the description can you can change the descriptions to, for example, `     description: pickup => ['health potion description', (pickup.power * healthPotionRestoreAmount).toString()],`

jdoleary (commented):
This is super fun to use!  One small additional change needed and then it'll be good to go!

src/cards/merge.ts
SoulMuncher: The simplest version I can think of is spawning a weaker/smaller Spellmason near the main or "merged into" target (Imagine a spellmason with split or split 2 on it, for example). Spawning it in the middle of the two players could cause some weird situations with terrain and whatnot.

SoulMuncher: Hmm, I do really like that idea from a gameplay standpoint, and it seems a little more intuitive / in-line with the regular spell behavior, but I wonder if it's worth giving up the tiny Spellmason? After showing a couple people, it feels like one of those "oh my gosh!" laughing moments between friends, and a feature people would adore and talk about, so I'm wondering if that's more valuable here? What do you think? Maybe we disable merging between ai/player units, and keep this in as a player-player easter egg?

jdoleary: I like the idea of the tiny spellmason but it doesn't make sense to me functionally.  Like I think it'll feel like a bug.  I wonder if there's another way we can introduce that.  Merge doesn't create any new units under any other circumstances.
Let's go with my last take: " it kills the other player (leaving their body) and gives their stats (current stats, not max stats) to the caster. "
I think this makes the most sense given the description of the spell.  And we can backburner the tiny spellmason idea, maybe we can make it an easter egg some other way

SoulMuncher: Done. One visual issue: Predictions don't show the stamina a player would gain by merging.

jdoleary:
bug: I target circle'd a group of 3 enemies (archer, archer, and golem), the merge animated a particle from the archer that disappeared into the golem (who nothing happened to).  You'll probably need to filter by unit type somewhere

src/graphics/PlanningView.ts
```diff
@@ -335,18 +335,17 @@ export function drawHealthBarAboveHead(unitIndex: number, underworld: Underworld
       // Don't draw healthbars for Doodads
       return;
     }
-    const predictionUnit = !underworld.unitsPrediction ? undefined : underworld.unitsPrediction[unitIndex];
+    const predictionUnit = u.predictionCopy;
```
jdoleary, line 338: Good catch!!

jdoleary (commented):
Looks great overall! Small bug with the animated particle and then it's good to go!

jdoleary (approved):

PR #642, opened 2024-04-17
Drown Variations

### LINKED ISSUE #539: Spell: Drown Variations
1. Damages all units that are submerged wherever they are
2. Huge percentage damage to targeted units submerged

@SoulMuncher thoughts?
I think drown is a rather blah spell atm, these upgrades could make it much more interesting

SoulMuncher:
Closes #539 

## TODO
- Blood Bath as a targeting spell instead?
- QA
- Balancing
- Descriptions
- FX


src/Underworld.ts
```diff
@@ -3443,6 +3443,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     }
   }
   // TODO - This should factor in fortify, debilitate, bloat explosions, etc.
+  // Any reason we dont fully simulate the enemy turn and store relevant results?
```
jdoleary, line 3446: I suspect the enemy turn would take too long to calculate at prediction speeds.  Due to all of them processing collisions and pathing etc

src/Underworld.ts
```diff
@@ -3472,7 +3473,7 @@ ${CardUI.cardListToImages(player.stats.longestSpell)}
     }
     switch (u.unitSubType) {
       case UnitSubType.MELEE:
-        const maxPathDistance = u.attackRange + u.stamina;
+        const maxPathDistance = u.attackRange + (Unit.canMove(u) ? u.stamina : 0);
```
jdoleary, line 3476: Just curious, was there a found bug that this fixes or did you just notice it was missing?  Just wondering under what circumstances this would matter, I don't recall prediction not working for frozen units for example

src/entity/Unit.ts
```diff
@@ -1350,47 +1356,49 @@ export function inRange(unit: IUnit, target: Vec2): boolean {
...
-  for (let unit of units) {
-    turnStartPromises.push(runTurnStartEvents(unit, underworld, prediction))
+  // Add mana to Player units
+  for (let unit of units.filter(u => u.unitType == UnitType.PLAYER_CONTROLLED && u.alive)) {
+    // Restore player to max mana at start of turn
+    // Let mana remain above max if it already is
+    // (due to other influences like mana potions, spells, perks, etc);
+    unit.mana = Math.max(unit.manaMax, unit.mana);
   }
-  await raceTimeout(5000, 'Turn Start Events did not resolve', Promise.all(turnStartPromises));
```
jdoleary, line 1358: What's the reasoning for the changes in this file?

jdoleary: More descriptive comments would be helpful.  This comment just says "Fix Turn Start/End organization " but that provides no context about why the change was made and what's improved.  Descriptive commit messages are especially helpful when looking back in time to see why something was changed long ago.

src/Underworld.ts
SoulMuncher: This fixed a prediction issue where a melee unit with dark tide would still show as attacking the player.

Freeze didn't have the same issue is because frozen units **can't act** at all and are skipped in the smart targeting function before this logic runs. Dark Tide units **can act** (i.e. archers shooting from the liquid), they just can't move, and whether or not they could move wasn't being checked here, only their stamina.

src/entity/Unit.ts
SoulMuncher: The initial problem was that dark tide and freeze were setting stamina to 0 in turnStart events, but units were immediately regenerating stamina *after* the turn start events finished. I thought about the ordering more, and the way I had written it previously didn't make as much sense, it was a mistake when I did the turn phase refactor. Afaik the only real effect of the ordering was undoing freeze's stamina change, but this didnt cause any issues in-game since freeze now uses canAct and that sort've overrides whether or not a unit can do things, so stamina at that point doesnt matter.

This was just cleanup for all of that it makes sense that units would generate stamina first (or players generate mana first), before we run a bunch of turn start logic that might then affect those stats.

SoulMuncher:
Anything else missing? Do you want any special vfx for either spell?

src/entity/Unit.ts
jdoleary: That makes sense! Thanks!

jdoleary (approved):

Issue #629, opened 2024-04-14
Soft lock with portal on + levels


jdoleary:
> Got soft locked out of the game. Exit Portal showed up on level 1+, round one. I took it because I could not beat round two, only to find that I am perpetually stuck in the game, without being able to cast spells. Had to abandon this run, losing my data on how well the run went because I technically didn't die. 

https://steamcommunity.com/app/1618380/discussions/0/7845908402042727539/

PR #653, opened 2024-04-22
ref: Allow purple portal to be used mid-game

jdoleary:
Refactor purple portals to always allow respawning.  The main motivation here is to resolve the soft-lock scenario where if a player enters a purple portal the game gets stuck if it does not progress.  Now purple portals will always allow you to respawn.  I've also changed it so purple portals have a very slim chance of appearing on their own now that they can be used as an item and made it so they get removed after one use. (no more exception for purple portals).
I also checked that if there aren't enough portals, more spawn for the other players when one player goes through one.

I think this change is cool, but I'm a little concerned about creating new corner-case bugs.  Curious to get your feedback.

Original description:
Rare purple portals allow players to respawn at a location of their choice. Consider any unspawned players as "inPortal"
Refactor some of enterPortal into resetPlayerForSpawn so that it can also be used in enterPortal
This also solves the soft lock bug: #629


jdoleary:
TODO: The end turn / portal logic needs to be better unified so there are no gaps for corner case bugs

PR #685, opened 2024-04-29
Merge Fixes

### LINKED ISSUE #681: Merging another player change
Merging another player should still kill them but not change their max stats (which ruins the run)
Instead, let's have it just transfer current stats (health instead of maxHealth, etc) 

Just for merging with other players

SoulMuncher:
closes #681

Players gain current stats instead of max stats when being merged into.


jdoleary (approved):
Works perfect

PR #716, opened 2024-05-11
Fix Camera Clamps

### LINKED ISSUE #715: New camera clamps too much when zoomed out
When zoomed out you often can't move the camera side to side, which feels really claustrophobic.

You should be able to "center" the camera over any point within the bounding box of the level, regardless of zoom, instead of restricting movement while the level is still on screen

SoulMuncher:
closes #715 

Allows manual camera movement when zoomed out, and simplifies code:
- While autofollowing, the outer edges of the camera will clamp to the map limits, showing as little empty space as possible without feeling uncomfortable
- While manually controlling the camera, the inner edges of the camera will clamp to the map limits, allowing you to move it freely while always keeping some part of the map on the screen (or barely off the screen) so players don't get lost

## TODO
Done


jdoleary (approved):

jdoleary:
Fixed in https://github.com/jdoleary/Spellmasons/commit/30db5e40a37548be25c1935bf57441c4830dc543

PR #717, opened 2024-05-12
All bolts play at once

### LINKED ISSUE #714: Bolt casts on many enemies triggers on each individually
...in sequence and can also take a long time to animate

SoulMuncher:
closes #714

Bolt targets are stored and then animated/damaged all at once, to fix an issue where it would animate on each individually and damage sequentially

## TODO
Done

### DIFF
--- a/src/cards/bolt.ts
+++ b/src/cards/bolt.ts
@@ -36,6 +36,7 @@ const spell: Spell = {
       if (length) {
         playDefaultSpellSFX(card, prediction);
       }
+      const affected: HasSpace[] = [];
       for (let i = 0; i < length; i++) {
         const target = targets[i];
         if (target) {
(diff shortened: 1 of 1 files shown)


jdoleary:
This changes how bolt does damage.  If you target circle two enemies near each other, the surrounding enemies take only 8 damage instead of 16.  Is this intentional?

SoulMuncher:
> This changes how bolt does damage. If you target circle two enemies near each other, the surrounding enemies take only 8 damage instead of 16. Is this intentional?

Yes. Only untargeted enemies could be hit by multiple bolts, so it seemed like a bug. Two enemies side by side in target circle + bolt would only take 8 damage, so that was the implementation I stuck with. It just means that a third enemy outside of the circle will continue to follow that pattern of no-stacking damage and take 8 as well, instead of 16.

As long as it's overlapping on everything or nothing, instead of some things, I think it'll feel good and look intentional. It would be reasonable to make the bolt overlap on all units (targeted ones included) if you wanted to do that instead.

jdoleary (approved):

PR #736, opened 2024-05-18
Fix: Burst quantity damage

SoulMuncher:
Fixes an issue where Burst's damage would be multiplied by quantity twice

## TODO
- Done


jdoleary (approved):
Good find!

PR #768, opened 2024-05-22
Implement Stamina Circles

### LINKED ISSUE #422: Stamina Circles
@ Dangleface
https://discord.com/channels/1032294536640200766/1202299187257942016

Inkbound released their Movement 2.0 update in 2023, which changed movement from using “Will” (AKA action points) to using a sort’ve dynamic stamina circle.

You can move up to a maximum distance away from your starting location each turn, and your stamina usage is only calculated and finalized when you take a game action such as attacking an enemy, at which point your remaining stamina creates a new max distance circle for you to move around in. Here’s the patch notes for the Movement 2.0 update

https://steamdb.info/patchnotes/11979511/

### LINKED ISSUE #745: Stamina Circle (refactor)
Instead of stamina just being spent when you walk, you have a stamina circle that you can move within. Any **action** taken within that circle caps your stamina for the turn at what it was when the action was taken. Also being force moved (say an ally casts force move on you) will also cap stamina.

I am undecided if taking potions should cap the stamina or not but for now let's try yes.

SoulMuncher:
closes #422 
closes #745 

Players can walk around as much as they want within [stamina] units of their starting point.

## TODO
- Check for other issues that can be closed
- Implement "locking" after spell action/force move/pickup
- QA single/multiplayer

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -52,7 +52,7 @@ import type { Vec2 } from "./jmath/Vec";
 import * as Vec from "./jmath/Vec";
 import Events from './Events';
 import { UnitSource, allUnits } from './entity/units';
-import { clearSpellEffectProjection, clearTints, drawHealthBarAboveHead, drawUnitMarker, isOutOfBounds, runPredictions, updatePlanningView } from './graphics/PlanningView';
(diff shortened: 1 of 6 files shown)


jdoleary:
I like it so far!
Found one bug where if i use the "teleport" spell it leaves behind the bright yellow circle and the dim yellow circle is next to me and I can't move.
It seems like it'd be helpful to see what my locked stamina will be if I have a spell queued and I hold "F".  So like if my temp stamina is "50%" but it's not locked in yet, when I'm moving I can see the transparent gold circle which is great but when I stop moving I can't see it anymore even if I hold "F"

PR #786, opened 2024-05-28
Plus Radius can now be cast on its own

### LINKED ISSUE #762: Plus Radius behaviour is inconsistent with urns
Affecting an urn with any spell that contains plus radius will permanently increase the explosion radius of the urn (which may or may not be a feature)

The behavior is inconsistent though, since you can't just cast plus radius on the urn directly. You can increase radius by doing something like [plus radius, poison, purify] but you can't just cast [plus radius] by itself

SoulMuncher:
closes #762 

Which lets players increase Urn Radius without needing another spell

## TODO
- Is there a clean way to improve refund logic? (I.E. should [Plus Radius + Slash] refund plus radius since it has no effect on the spell?)
- Done otherwise

### DIFF
--- a/src/cards/plus_radius.ts
+++ b/src/cards/plus_radius.ts
@@ -14,7 +14,6 @@ const spell: Spell = {
     expenseScaling: 1,
     probability: probabilityMap[CardRarity.UNCOMMON],
     thumbnail: 'spellIconPlusRadius.png',
-    requiresFollowingCard: true,
     description: 'spell_plus_radius',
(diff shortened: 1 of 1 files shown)


jdoleary:
> Is there a clean way to improve refund logic? (I.E. should [Plus Radius + Slash] refund plus radius since it has no effect on the spell?)
I say this is on the player, if they use it with no effect then they'll just have to learn

jdoleary (approved):

PR #788, opened 2024-05-29
Spell Damage / Functional Player Damage

### LINKED ISSUE #578: Damage as % of Base
TL;DR:
Damage bonuses in games are fun, but not easily doable in Spellmasons at the moment. If spells scaled off of the player's damage stat, we would be able to add damage bonuses/penalties to them. This means already existing spells like empower/enfeeble would have a use case on players, and we could create even more new content surrounding damage increases/decreases, which is really nice. Should have a very low time cost.

> @SoulMuncher 03/20/2024 3:33 PM
> Kind've on that topic, I've wondered if spells should take from a core damage stat anyway. Say spellmasons have a base damage of 20 that can be increased/decreased by various effects. Slash would do 100% of the character's damage stat, Arrows would do 50%, Heavy Slash would do 200%, etc. If the damage stat doubles, all of the spells deal double damage. 
> 
> The benefit is that we actually have a way to give damage bonuses to players. And it wouldn't be a flat increase that ends up OP on Arrow and un-noticeable on Megaslash, it would be inherently balanced / proportional. Things like Empower can apply to any unit, including Spellmasons, without any weird edge cases. We can give damage bonuses via Runes, power-circle auras, spells, support units, boss mechanics (I'm thinking of some cool stuff), meta progression, even items or something if you go crazy with it. Damage could be a stat option when allocating points, which is kind've interesting to think about.
> 
> Not sure if there's a downside to doing it this way. It's a pretty common system so I don't think it'd be unintuitive, plus it would only really apply if the player has run into one of these damage changing effects, such as casting empower on themselves. Even then, the tooltip can use the calculated damage value, and predictions will handle everything else just fine.
> 
> Would probably only take ~3 hours to implement it, or about an hour to push a small example with Slash / Arrow / Empower if you're not sold on it.

SoulMuncher:
Closes #578 

Spells now scale off of the player unit's damage stat via a damage multiplier, instead of having a constant damage value.

## TODO
- Update mods?
- Consider adding "Damage" stat option to player perk selection
- Consider rebalancing the Goru (since its bone shrapnel now scales with damage)
- QA Multiplayer

### DIFF
--- a/src/cards/add_heal.ts
+++ b/src/cards/add_heal.ts
@@ -2,9 +2,10 @@ import { CardCategory } from '../types/commonTypes';
 import { Spell } from './index';
 import { CardRarity, probabilityMap } from '../types/commonTypes';
 import { healSfx, healUnits } from '../effects/heal';
+import { GetSpellDamage } from '../entity/Unit';
 
 export const healCardId = 'heal';
-const healAmount = 30;
+const healMult = 1.5;
 
 const spell: Spell = {
   card: {
@@ -18,9 +19,9 @@ const spell: Spell = {
     probability: probabilityMap[CardRarity.COMMON],
     thumbnail: 'spellIconHeal.png',
     animationPath: 'spell-effects/potionPickup',
-    description: ['spell_heal', healAmount.toString()],
+    description: ['spell_heal', GetSpellDamage(undefined, healMult).toString()],
     effect: async (state, card, quantity, underworld, prediction) => {
-      await healUnits(state.targetedUnits, healAmount * quantity, state.casterUnit, underworld, prediction, state);
+      await healUnits(state.targetedUnits, GetSpellDamage(state.casterUnit.damage, healMult) * quantity, state.casterUnit, underworld, prediction, state);
       return state;
     },
   },
--- a/src/cards/arrow.ts
+++ b/src/cards/arrow.ts
@@ -52,7 +52,7 @@ const spell: Spell = {
     }
   }
 };
-export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPierce: boolean = false) {
+export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPierce: boolean = false, extraIgnoreUnitIds: number[] = []) {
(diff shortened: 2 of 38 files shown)


SoulMuncher:
@jdoleary 

Won't run on headless:
```
TypeError: Unit.GetSpellDamage is not a function
    at Object.<anonymous> (C:\Users\Jamis\Desktop\Spellmasons\headless-server-build\src\cards\poison.js:104:44)
    at Module._compile (node:internal/modules/cjs/loader:1369:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)
    at Module.load (node:internal/modules/cjs/loader:1206:32)
    at Module._load (node:internal/modules/cjs/loader:1022:12)
    at Module.require (node:internal/modules/cjs/loader:1231:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (C:\Users\Jamis\Desktop\Spellmasons\headless-server-build\src\Upgrade.js:39:18)
    at Module._compile (node:internal/modules/cjs/loader:1369:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1427:10)```

jdoleary:
This change is too wide-reaching and I am concerned about the impact of the balance of the game.  We may revisit this idea later but for now I'm just modifying empower so that it's clear it doesn't impact players.

* 3d46eacb (HEAD -> master, origin/master) fix: Prevent empower from targeting players
* 7e1f9e4d i18n: Clarify that Empower does not affect players

jdoleary:
Reopening.  Empower needs to be reverted but we'll keep the rest.

SoulMuncher:
@jdoleary 
Two things to do before we can pull the PR
- Update Enfeeble description to match Empower description
- Fix card description variables (description variables should stay up to date with player stats, instead of only being initialized once at game startup)

Spell description variables already factor in the client player's stats, I.E. `Slash deals X damage`, but the problem is that they initialize once on startup and never update after that, meaning any runtime change isn't reflected. They are locked at showing the base values. Do you have a good idea/fix for this? Is there a quick way to "recalc" or update the description whenever its accessed? Maybe if we change it to some kind of `=>` function?

Issue #382, opened 2024-01-14
Steam Achievements!


SoulMuncher:
# Achievements
Need flags that can be set by some game events, reset per-game / per-level, and checked for achievements.

### LevelComplete()
- Clear the final stage (12)
- Clear the final stage (12) in a multiplayer game
- Complete the game under X conditions
  - with [category] spells disabled in run options
  - without obtaining any runes
  - with all stat points allocated to good looks
  - without taking damage
  - with certain challenge settings enabled
- Complete a level after you've died
- Finish a level with <10 hp remaining

### Cast Spell() / Creative
- Cast a spell that...
  - Costs X mana
  - Deals X damage/healing
  - Kills X targets
  - ...
- Shoot 100+ arrows with one spell
- Purify 5+ curses at once
- Attempt to cast capture soul on a purple portal
- Inflict impending doom on an enemy
- Capture the Deathmason's soul
- Create a copy of yourself
- Figure out how to use targetInjured as the first card in the first spell cast on a level
- Freeze yourself

### Silly
- Click the join discord button
- Scroll to the bottom of the legal document
- Don't skip the tutorial (if the player skips the tutorial, they can acquire this later the via reset tutorial button)

### Metagamey
- Unlock X/Y/Z% of the codex
- Discover all of the spells
- Unlock all achievements

Other:
- Have over 1000 health, stamina, and mana at once
- Kill a monster using the environment (liquid/urn/trap/impact damage)
- Kill the Deathmason twice in one run (enable natural boss spawning in the + levels with budget rework?)

SoulMuncher:
https://discord.com/channels/1032294536640200766/1199444010838786048

jdoleary:
> meme_man
> OP
>  — Yesterday at 6:54 PM
> "thats a cold move" when you freeze yourself

jdoleary:
@SoulMuncher 
Setting up an achievement system might be tricky to do it modularly and decoupled.  I'd say each achievement could take an event callback and would return true if achieved.
So we could have a list of achievement-specific events like: onEndgame, onPlayerDeath, onChooseUpgrade, etc and add to that as needed
to get started I would compile a large list of possible achievements and organize them by triggering event before designing the system that would handle the achievements.
If you make the achievement event system I can handle hooking it up to the steam backend

PR #808, opened 2024-06-07
Achievements / Statistics

SoulMuncher:
progress on #382 

Currently contains:
- First draft commit of achievement/statistics system to outline approach to code structure

## TODO
- Discuss draft with @jdoleary 
- Make Todo's

### DIFF
--- a/src/Achievements.ts
+++ b/src/Achievements.ts
@@ -1,0 +1,84 @@
+import * as GameStatistics from "./GameStatistics";
+import Underworld from "./Underworld";
+import { LAST_LEVEL_INDEX } from "./config";
+
+//
+
+export const allAchievements: { [key: string]: IAchievement } = {};
+
+interface IAchievement {
+  id: string,
+  description: string,
+  unlocked: boolean,
+}
+
+//
+
+const achievement_CompleteTheGame: IAchievement = {
+  id: "Complete the game",
+  description: "Complete the first 12 levels",
+  unlocked: false,
+}
+const achievement_CompleteGameNoDamageTaken: IAchievement = {
+  id: "Untouchable",
+  description: "Complete the first 12 levels without taking any damage",
+  unlocked: false,
+}
+const achievement_ArrowRain: IAchievement = {
+  id: "Arrow Rain",
+  description: "Shoot 100 or more arrows with one spell",
+  unlocked: false,
+}
+
+//
+
+// ID's and registry maybe not needed?
+function registerAchievement(achievement: IAchievement) {
+  allAchievements[achievement.id] = achievement;
+}
+
+// ID's and registry maybe not needed?
+export function registerAllAchievements() {
+  registerAchievement(achievement_CompleteTheGame);
+  registerAchievement(achievement_CompleteGameNoDamageTaken);
+  registerAchievement(achievement_ArrowRain);
+  console.log("Registered achievements!", allAchievements);
+}
+
+//
+
+export function UnlockAchievement(achievement: IAchievement) {
+  if (!achievement.unlocked) {
+    achievement.unlocked = true;
+    console.log("Achievement unlocked!", achievement)
+  } else {
+    console.log("Achievement is already unlocked.", achievement)
+  }
+}
+
+// ID's and registry maybe not needed?
+export function UnlockAchievementById(id: string) {
+  const achievement = allAchievements[id];
+  if (achievement) {
+    UnlockAchievement(achievement)
+  }
+}
+
+export function UnlockEvent_CastCards(underworld: Underworld) {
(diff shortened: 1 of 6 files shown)


src/GameStatistics.ts
```diff
@@ -0,0 +1,74 @@
...
+
+//
+
+export interface IStatistics {
+  cardsCast: number;
+  myPlayerArrowsFired: number;
+  myPlayerDamageTaken: number;
+}
+
+// How can we initialize these without manually typing everything out 3+ times?
```
jdoleary, line 12: Use can use `Object.assign({}, { cardsCast: 0, myPlayerArrowsFired: 0, myPlayerDamageTaken: 0 })` to clone the object.  (note it's a shallow clone but that should be fine since this object only has primitives and no nested objects.

jdoleary (commented):
Looking great!

src/Achievements.ts
```diff
@@ -0,0 +1,139 @@
...
+export const achievement_Doomsayer: IAchievement = {
+  id: "Doomsayer",
+  description: "Curse an enemy with 'Impending Doom'",
+  unlocked: false,
+}
+const achievement_100Percent: IAchievement = {
+  id: "100 Percent",
+  description: "Earn all other achievements",
+  unlocked: false,
+}
```
jdoleary, line 62: These are awesome!

jdoleary:
What's the status of this? Will it be ready soon?

jdoleary:
I think the achievements we have so far is a good start, I'd say let's wrap up the outstanding TODOs and get this shippable for the next update. Sound good?  We can add more achievements as we go

src/network/networkHandler.ts
```diff
@@ -1054,6 +1055,13 @@ async function handleLoadGameState(payload: {
   underworld.enemiesKilled = loadedGameState.enemiesKilled;
   underworld.activeMods = loadedGameState.activeMods;
   underworld.statCalamities = loadedGameState.statCalamities || [];
+  underworld.allowAchievements = loadedGameState.allowAchievements;
+  Object.assign(GameStatistics.underworldStats, loadedGameState.underworldStatsForSerialization);
+  if (!(underworld.players.length > 1)) {
```
jdoleary, line 1060: This `if` statement is hard to understand at first glance.  How about `underworld.players.length == 1`?

src/cards/purify.ts
```diff
@@ -39,12 +41,18 @@ const spell: Spell = {
...
+  let cursesRemoved = 0;
   for (let [modifier, modifierProperties] of Object.entries(unit.modifiers)) {
     if (modifierProperties.isCurse) {
       Unit.removeModifier(unit, modifier, underworld);
+      cursesRemoved += 1;
+      GameStatistics.trackCursePurified({ unit, sourceUnit }, underworld, prediction);
     }
   }
+
+  if (!prediction && sourceUnit == globalThis.player?.unit && cursesRemoved >= 5) {
```
jdoleary, line 54: Rather than using the cursesRemoved variable on 3 lines, let's just do `unit.modifiers.filter(m => m.isCurse).length` here.  it keeps the logic more concise.

src/GameStatistics.ts
```diff
@@ -0,0 +1,276 @@
...
+  EmptyStatistics(), // 2 - Level
+  EmptyStatistics(), // 3 - Spell
+];
+export interface IStatistics {
+  unitDeaths: number;
+  myPlayerDamageTaken: number;
+  myPlayerDeaths: number;
+  cardsCast: number;
+  myPlayerArrowsFired: number;
+  myPlayerCursesPurified: number;
```
jdoleary, line 60: We're tracking `myPlayerCursesPurified` but we're not querying it anywhere, let's remove it and its corresponding function

src/cards/arrow.ts
```diff
@@ -98,6 +99,8 @@ export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPi
             collideFnKey
           }, underworld, prediction);
 
+          GameStatistics.trackArrowFired({ sourceUnit: state.casterUnit }, underworld, prediction);
```
jdoleary, line 102: I want to rethink how this is being calculated.  The CPU is doing a ton of extra work here.  For every arrow it's creating a new array (via slice) containing 3 existing Stats objects and then incrementing all of them, and then that's done over and over again for each arrow.  Now this may not be noticeable perf-wise, but this is the kind of thing that could become a problem later if it's changed in some way given that it's invoked so much for something that could just be a single if statement somewhere.

Let's just add `arrowsFired` to effectState.aggregator, increment that inside arrowEffect and then check it at the end of the spell.

Issue #818, opened 2024-06-17
Freeze sprite missing on load (multiplayer)


jdoleary:
Start multiplayer game,
freeze a unit
join with second player
note that second player does not see freeze sprite

PR #823, opened 2024-06-18
Modifier.Init wasn't using prediction in load unit

SoulMuncher:
@jdoleary 
Just stumbled across this line of code and wasn't sure if it was intentional and why. Not sure what issues it might affect. Maybe stuff like #818 ?


jdoleary:
Good catch!

Issue #820, opened 2024-06-17
Add events needed to implement runes


jdoleary:

jdoleary:
@SoulMuncher I'm gunna take this one unless you've already started on it?

SoulMuncher:
@jdoleary
OnKill and OnPickup are included in #918 👍

PR #828, opened 2024-06-19
Refactor events

jdoleary:
ref: Collect all events in a single array
    
    This removes several footguns that occurred when
    adding a new type of event.
    Much less boilerplate now that all events are handled in
    the same array and since it always has to check the eventSource
    in order to invoke the function it is okay that different
    event types are stored in the same array.
    
    One potential (I think unrealized) problem with this approach
    is if some modifier needs to selectively remove one event type
    but not another - however this can now just be done with
    different event ids. And I don't think this currently happens in the
    codebase so I don't see it as a problem.
    
    This commit paves the way for Runes which will be very event heavy
    and will add lots of new event source types.


jdoleary:
This lays the groundwork for #820

SoulMuncher:
I can't think of any places where we do or would selectively remove one event type of a given id and since the event source is already checked, it makes sense to put it all of the events into one array. This looks good to me.

PR #815, opened 2024-06-13
Reroll Omit Enhancements

SoulMuncher:
closes #802 
closes #767

Ensures rerolled cards won't reappear within a single upgrade selection

## TODO
- Done

PR #832, opened 2024-06-22
Level Up Fix

### LINKED ISSUE #790: Why don't you level up past 98?

SoulMuncher:
closes #790 
also fixes a bug where you would only level up once upon killing an enemy, regardless of how many levels you were eligible to gain

## TODO
- Done

### DIFF
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -300,17 +300,17 @@ export default class Underworld {
       console.log('HUD-hidden: Skipping dropping scroll pickup')
       return;
     }
-    const numberOfEnemiesKilledNeededForNextDrop = this.getNumberOfEnemyKillsNeededForNextLevelUp();
-    if (numberOfEnemiesKilledNeededForNextDrop <= this.enemiesKilled) {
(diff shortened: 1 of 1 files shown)


jdoleary:
How was this tested?
I'm able to level up multiple levels in master build without this change already

SoulMuncher:
From what I tested on master, if you type devUnderworld.enemiesKilled = 1000, and then kill an enemy, you'll only level up once.
This _should_ never be a problem in game, since exp is 1:1 with enemies killed, but we've seen a couple exp bugs and this will handle those more smoothly. For dev purposes, it should also speed up testing things like #815 in the future

jdoleary:
If you want to test leveling up you could invoke devUnderworld.reportEnemyKilled() in a loop.  Setting enemiesKilled in console is not how it works in game so that's not an actual reproduction of game behavior.

SoulMuncher:
@jdoleary EXP Bug example: https://discord.com/channels/1032294536640200766/1236220753217847367/1241836668789592176

Issue #759, opened 2024-05-21
Souls as a resource


jdoleary:
new resource "Souls" that you would have to spend to use Soul magic: resurrect, summon, etc
I'm thinking they would be finite, non-renewable, persist between levels and you earn more by killing "original life" mobs

Discussion:
SoulMuncher — Today at 4:30 PM
Hm
I love the idea mechanically and I like how it balances some of those bigger spells
JordanO — Today at 4:30 PM
I think it would be really interesting to think about "saving for future levels"
SoulMuncher — Today at 4:31 PM
Ahh.. idk about that
like resources carrying over?
JordanO — Today at 4:31 PM
yeah
SoulMuncher — Today at 4:31 PM
One of my favorite things about spellmasons is how it doesn't do that
JordanO — Today at 4:31 PM
hmm ok
so you'd start each level with 0 souls?
SoulMuncher — Today at 4:31 PM
Yeah I think so
JordanO — Today at 4:31 PM
that would make necro build hard
or maybe a small number?
SoulMuncher — Today at 4:32 PM
Necro class starts with 3 maybe?
JordanO — Today at 4:32 PM
maybe that could be upgraded: "Starting Souls"
SoulMuncher — Today at 4:32 PM
Oh true
rune idea!

jdoleary:
Closing for now, cool idea but I think the game functions well without it

PR #844, opened 2024-07-06
Souls as a Resource

SoulMuncher:
#759 

### PROTOTYPE - FOR TESTING ONLY

- Adds a new resource called "Souls"
- Spells in the soul category cost 1 soul per quantity, in addition to their mana cost
- Players can acquire souls by killing originalLife units
- Souls are currently displayed in the selected unit info panel
- Start each level with 0 souls (Necromancer starts each level with 3 souls instead)

## TODO
- Test and discuss

### DIFF
--- a/src/graphics/PlanningView.ts
+++ b/src/graphics/PlanningView.ts
@@ -889,6 +889,7 @@ ${globalThis.selectedUnit.faction == Faction.ALLY ? '🤝' : '⚔️️'} ${i18n
 🎯 ${globalThis.selectedUnit.attackRange} ${i18n(['attack range'])}` : ''}
 ❤️ ${globalThis.selectedUnit.health}/${globalThis.selectedUnit.healthMax} ${i18n(['health capacity'])}
 🔵 ${globalThis.selectedUnit.mana}/${globalThis.selectedUnit.manaMax} + ${globalThis.selectedUnit.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}
+👻 ${globalThis.selectedUnit.souls} ${i18n('Souls')}
 ${extraText}
 ${playerSpecificInfo}
       `;
--- a/src/Underworld.ts
+++ b/src/Underworld.ts
@@ -46,7 +46,7 @@ import {
   cacheBlood,
 } from './graphics/PixiUtils';
 import floatingText, { queueCenteredFloatingText, warnNoMoreSpellsToChoose } from './graphics/FloatingText';
-import { UnitType, Faction, UnitSubType, GameMode } from './types/commonTypes';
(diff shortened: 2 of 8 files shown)


src/graphics/PlanningView.ts
```diff
@@ -889,6 +889,7 @@ ${globalThis.selectedUnit.faction == Faction.ALLY ? '🤝' : '⚔️️'} ${i18n
 🎯 ${globalThis.selectedUnit.attackRange} ${i18n(['attack range'])}` : ''}
 ❤️ ${globalThis.selectedUnit.health}/${globalThis.selectedUnit.healthMax} ${i18n(['health capacity'])}
 🔵 ${globalThis.selectedUnit.mana}/${globalThis.selectedUnit.manaMax} + ${globalThis.selectedUnit.manaPerTurn} ${i18n('Mana')} ${i18n('per turn')}
+👻 ${globalThis.selectedUnit.souls} ${i18n('Souls')}
```
jdoleary, line 892: haha YES!

jdoleary (commented):
Wow this is great already!

We'll have to consider how merge and split work with a unit's souls.
Resurrect currently only uses one soul, and I'm wondering if we should do 1 soul per unit resurrected.
Also should souls be recyclable?  Like if one of your minions gets killed should you get the soul back or have a way to re-harvest the soul?  I'm thinking there should be a new spell that allows you to harvest souls of the dead so you don't just get them automatically.  That way in multiplayer if one player is going necro build, they could be the one to get all the souls.

Looking great so far!

SoulMuncher:
@jdoleary 
> Resurrect currently only uses one soul, and I'm wondering if we should do 1 soul per unit resurrected.

Balance-wise, yes! But will it be intuitive? The units have a resurrect icon over their head, so in theory the player should understand what's going on. Yeah, that sounds good.

> Also should souls be recyclable? Like if one of your minions gets killed should you get the soul back or have a way to re-harvest the soul?

As a baseline, I don't think so. You get one soul, you spend it on a summon, and that summon will create value until the end of its lifetime. Being able to summon it again "for free" would be like getting a free slash this turn because you used slash last turn. BUT, as a rune idea or costly spell? Totally. I like the idea of Souls being a little sparse/difficult to collect by default, but with accessible upgrades like starting soul count/additional harvesting/recycling/etc. Good for game balance, good for content, good for player builds and diversifying summon playstyles

> I'm thinking there should be a new spell that allows you to harvest souls of the dead so you don't just get them automatically. That way in multiplayer if one player is going necro build, they could be the one to get all the souls.

Yeah maybe. For "default behavior" I figured whoever gets the kill gets the soul, but I didn't get that far with the prototype. The code for it seemed somewhat complicated when you factor in summoned units and modifiers and whatnot. Those don't track a "source" back to the player caster.

PR #861, opened 2024-07-12
Add Shield Regen

SoulMuncher:
New Rune/Modifier:

Shield Regen
Regenerates (quantity) shield at the start of each turn

- This PR also makes some changes to how the Shield modifier works: 1 stack/quantity of shield now = 1 damage blocked instead of 30, and the spell has been changed to account for that, now adding 30 quantity per cast

## TODO
- Done


SoulMuncher:

PR #862, opened 2024-07-12
Add Revitalize

SoulMuncher:
New Rune/Modifier:

Revitalize
Increases incoming healing by (quantity)%

- This PR also makes some changes to debilitate quantity, such that each quantity of debilitate increases incoming damage by 1%. The spell has been adjusted to reflect this change
- Also fixes a bug + formula issue where each debilitate after the first would give +200% incoming damage instead of +100%

## TODO
- Done


SoulMuncher:

PR #863, opened 2024-07-12
Add Armor

SoulMuncher:
New Rune/Modifier:

Armor
Reduces incoming damage by (quantity)

Currently does not reduce incoming damage below 1

## TODO:
- It might be a good idea to implement modifier "order", so we can ensure modifiers are applying the way we and players expect/want them to. Armor should always apply after all other increases/reductions (debilitate)


SoulMuncher:

PR #864, opened 2024-07-12
Add Thorns

SoulMuncher:
New Rune/Modifier:

Thorns
Deals (quantity) damage to an attacker when taking damage

## TODO:
- Bug: If a unit with thorns damages a unit with thorns (itself or other) it will cause a chain reaction until one unit dies.
  - We will have to decide exactly what sources trigger what effects and how/why in a way that is intuitive. Will thorns trigger another thorns, or apply onHit poison/lifesteal? A lot of games differentiate between a "hit" and "dealing damage", where a "hit" can come from a spell like slash, and "damage" can come from thorns or damage over time.


SoulMuncher:

PR #865, opened 2024-07-13
Add On-Hit Poison

SoulMuncher:
New Rune/Modifier:

On Hit Poison
Applies (quantity) poison to a unit when hitting it

- This PR also makes some changes to poison quantity, such that each quantity of poison deals 1 damage per turn, making it more modular and compatible with the rune system. The spell has been adjusted to reflect this change

## TODO
- Better name?


SoulMuncher:

PR #868, opened 2024-07-13
New Runes

SoulMuncher:
#### <ins>New Modifiers/Runes:</ins>
- Armor - Reduces incoming damage by (quantity) - Currently does not reduce incoming damage below 1
- Health Regen - Regenerates (quantity) health at the start of each turn
- Mana Barrier - Damage is taken from remaining mana before health
- Overheal - Grants healing over max as shield at (quantity)% effectiveness
- On Hit Healing - Restores (quantity) health per hit
- On Hit Poison - Applies (quantity) poison to a unit when hitting it
- Revitalize - Increases incoming healing by (quantity)%
- Self Invulnerability - Grants invulnerability to one's own damage (Ex. explosives)
- Shield Regen - Regenerates (quantity) shield at the start of each turn
- Thorns - Deals (quantity) damage to an attacker when taking damage
- **Many More (WIP)**

#### <ins>Other:</ins>
- Changes to existing modifiers [Shield, Poison, Debilitate] such that each stack of the modifier results in 1 [Damage blocked / Damage per turn / Percent increase to incoming damage] respectively, allowing these modifiers to be added with specific values, rather than in multiples of [30, 20, 200] as their spells apply
- Fixes a bug + formula issue where each Debilitate after the first would give +200% incoming damage instead of +100%

## Important Discussion
- Order of operations: We should have a consistent modifier processing "order". Example order for the OnTakeDamage event:
  - Percent incoming damage/healing (Debilitate/Healing Effectiveness)
  - Flat incoming damage/healing (Armor)
  - Special incoming damage/healing
    - Blood Curse (In this example, Debilitate/Armor wouldn't affect the damage taken from heals while Blood Cursed, but invulnerability/thorns still trigger. Ideally, we'd have separate onTakeHeal and onTakeDamage events, so that Blood Curse could start a new OnTakeDamage event chain with TakeDamage(-amount), allowing relevant effects to apply to the new damage instance)
    - Invulnerability
  - Reactive effects ("If you would die/overheal", Thorns)
  
- Effect Sources and Triggers: What sources trigger what effects?
  - Many games differentiate between a "hit" and "dealing damage", where a "hit" can come from a spell like slash, and "damage" can come from poisons or other effects.
  - In our case, we need to consider what procs what and why, as well as how to put our design solution into code.
    - If Unit A takes thorns damage from Unit B, is the thorns damage reduced by Unit A's armor?
    - If Unit B has onHit poison or lifesteal, should their thorns damage trigger either effect?
    - If Unit A also has thorns, does it bounce back to Unit B? Does that then bounce damage back to Unit A and so on?
    - Can onHit poison/lifesteal be triggered via poison or summons
  - Say I restore 20 health OnKill:
    - Should it trigger if I kill an ally
    - Should it trigger on non-original units (Units that have been split/cloned/summoned)
    - Should it trigger if an enemy dies to liquid after I've pushed them
    - Should it trigger if my summon kills an enemy
    - Should it trigger if a poison applied by a summon of my summon kills an enemy
    - Should it trigger if a player ally kills an enemy
    - If p1 bloats a golem, p2 pushes the golem next to an archer, p2 kills the golem, and then the archer dies to bloat, who gets the kill for the archer?
    - If p1 and p2 both apply 20 poison to an archer and it dies on the next tick of poison damage, who gets the kill?
- Persistent Modifiers and Removal Events:
  - If I receive the Health Regen or Poison modifier from a spell, the modifier gets removed upon death. This should be the case for runes as well, but there needs to be an additional handler that reapplies those modifiers to ensure they last between levels and upon revival if needed. If a Rune gives me Health Regen, I should always have that health regen while alive.

## TODO
- Improve names, descriptions, etc.
- Cleanup
- QA

__
**~~TODO~~ (FUTURE PR)**
- Extra Pierce/Bounce modifier (Requires arrow PR)
- Modifier order
- Maybe better OnHit handling?
- OnKill event


src/cards/debilitate.ts
```diff
@@ -41,14 +41,16 @@ const spell: Spell = {
...
-      if (amount > 0) {
-        return amount * proportion * quantity;
-      } else {
-        // Do not magnify negative damage (which is healing)
-        return amount;
+      const modifier = unit.modifiers[id];
+      if (modifier) {
+        // Will only increase damage (doesn't affect incoming healing)
+        if (amount > 0) {
+          // Each quantity = 1% damage boost
```
jdoleary, line 48: 1%? not 100%?

src/cards/debilitate.ts
```diff
@@ -57,5 +59,20 @@ function add(unit: Unit.IUnit, _underworld: Underworld, _prediction: boolean, qu
...
   });
+
+  updateTooltip(unit);
+}
+
+function updateTooltip(unit: Unit.IUnit) {
+  const modifier = unit.modifiers[id];
+  if (modifier) {
+    // Set tooltip:
+    modifier.tooltip = `${CalcMult(modifier.quantity)}x ${i18n('Incoming')} ${i18n('Damage')}`;
```
jdoleary, line 70: Great addition!

src/modifierOnHitPoison.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+  });
+  registerEvents(onHitPoisonId, {
+    onDealDamage: (damageDealer: Unit.IUnit, amount: number, underworld: Underworld, prediction: boolean, damageReciever?: Unit.IUnit) => {
+      const modifier = damageDealer.modifiers[onHitPoisonId];
+      if (modifier) {
+        // Poison is only applied if damage is dealt, not when healing
+        if (damageReciever && amount > 0) {
+          // Apply poison to the damage reciever
+          Unit.addModifier(damageReciever, poisonCardId, underworld, prediction, modifier.quantity, { sourceUnitId: damageDealer.id });
```
jdoleary, line 29: I love this rune.  I think we could start to add some interesting spells that affect "Poisoned units" like how we do with drown.

src/modifierSelfImmunity.ts
```diff
@@ -0,0 +1,42 @@
...
+      }
+    }
+  });
+  registerEvents(selfInvulnerabilityId, {
+    onTakeDamage: (unit: Unit.IUnit, amount: number, underworld: Underworld, prediction: boolean, damageDealer?: Unit.IUnit) => {
+      const modifier = unit.modifiers[selfInvulnerabilityId];
+      if (modifier) {
+        // If the source of incoming damage is myself, negate it
+        if (damageDealer == unit && amount > 0) {
+          amount = 0;
```
jdoleary, line 27: We try to "show" when a modifier takes effect (via VFX or floating text) so there isn't any mystery as to why something happened.

src/modifierThorns.ts
```diff
@@ -0,0 +1,48 @@
...
+      const modifier = unit.modifiers[thornsId];
+      if (modifier) {
+        // Thorns will not deal damage if we are being healed
+        if (damageDealer && amount > 0) {
+          // Deal flat damage to the attacker
+          Unit.takeDamage({
+            unit: damageDealer,
+            amount: modifier.quantity,
+            sourceUnit: unit,
+          }, underworld, prediction);
```
jdoleary, line 32: Maybe we add an optional param to skip the onTakeDamage event so that we can avoid a thorns loop here?

jdoleary (commented):
Looks great so far! I'll have to do some more QA before merging

src/cards/debilitate.ts
SoulMuncher: Yes! Just note that this comment is referring to modifier quantity and not spell quantity: Each cast of the debilitate spell still increases incoming damage by 100% (but does so by adding 100 stacks of the modifier). Doing it this way just means it's possible to make an enemy take +x% damage where x isn't a multiple of 100. So we could make a rune that applies say, 25% debilitate instead of 100%, if 100% is too large of a number for the conditions of the rune to trigger

src/modifierOnHitPoison.ts
SoulMuncher: Yesss, that would be awesome! For common mechanics like poison and shield, I think it would be cool to have runes and modifiers that give you new ways of applying the effect, new reasons to apply the effect, and new interactions/synergies. 

It would be cool to see plague builds that help you apply and spread poison, but also give your poisons fun secondary effects like "your poisons now slow enemies"

A self-sustained thorns+lifesteal build that makes you near invulnerable to most regular/low level enemies.
A singletarget rend build that stacks more and more damage as you focus on one enemy, and grants some kind of onKill bounty reward or heal or something.
Crazy things like that

src/modifierThorns.ts
SoulMuncher: Maybe, but that would also allow thorns to bypass fortify, armor, invulnerability, debilitate, etc. I think our answer has to be more nuanced than that for some effects.

- We could differentiate between "on hit" and "on damage" events, such that onDamage events don't proc reactive onHit effects like thorns, but still check for the non-reactive effects like armor? We may still need to add a couple optional parameters on top of that as we go, like isDamageModifiable or something.
- Some games will keep track of what reactive effects have triggered from one "chain" of damage events and prevent them from triggering more than once.

When we find a modular solution, we should use it to replace the old Soul Bind logic as well.

SoulMuncher: Maybe "isHit" could be a boolean/optional parameter in the onTakeDamage event

jdoleary: Hmm, I don't want to over complicate it but I see what you mean.  Rather than adding extra events, what if we have a greater takeDamage callback that triggers everything (events and all) and then a secret inner function that just handles non-reactive stuff that is named such that it is obviously not meant to be used outside of special occasions.  Then SoulBind and Thorns could use the inner one

src/modifierOnHitPoison.ts
jdoleary: Love this. Can you add some of these to the runes list spreadsheet please?

JustinWick:
Apologies if this is the wrong forum, but if there's `healthRegen` and `shieldRegen`, why not `manaRegen`? It's less powerful than increasing max mana (it doesn't change things that scale off max mana) so it could be balanced to be "punchy" in its intended effects without being hard to balance with unwanted synergies. It's also just a cool effect that some cards might tack on: imagine a `Prismatic Blessing` card that gives `+10 Health Regen, +10 Shield Regen, +10 Mana Regen`.

JustinWick:
Unrelated, I find this deeply disturbing w/ regards to poison:

> Don't take damage on prediction because it is confusing for people to see the prediction damage that poison will do, they assume prediction damage is only from their direct cast, not including the start of the next turn.

For me, the fun in this game is being handed a problem and a bunch of tools and being asked to find some clever way to use those tools. The beauty of this as a video game, as opposed to a board game, is that the computer is doing all of the work of showing you what's about to happen while you're planning. Anything that interferes with that is an instant fun-killer, IMHO. I want to be able to put together that insane clutch play, and if poison damage should affect my decisionmaking it should be reflected in the prediction mechanism.

I think I've seen other games show it in the health bar of an enemy as its own segment for "you didn't do this direct damage but the unit will be taking it". I strongly suggest doing something like that here. If poison wrecks my ability to trust predictions I will avoid those cards despite all the interesting possibilities.

jdoleary:
> Apologies if this is the wrong forum, but if there's `healthRegen` and `shieldRegen`, why not `manaRegen`? It's less powerful than increasing max mana (it doesn't change things that scale off max mana) so it could be balanced to be "punchy" in its intended effects without being hard to balance with unwanted synergies. It's also just a cool effect that some cards might tack on: imagine a `Prismatic Blessing` card that gives `+10 Health Regen, +10 Shield Regen, +10 Mana Regen`.

So are you envisioning that this would result in overflow mana every turn?  For the players at least, since mana regen is always equal to max mana, there's actually no functional difference.  The idea is that max mana shows you how much mana you'll get every turn, but overflow mana is always allowed from spells or potions and so there's no cap.

jdoleary:
> Unrelated, I find this deeply disturbing w/ regards to poison:
> 
> > Don't take damage on prediction because it is confusing for people to see the prediction damage that poison will do, they assume prediction damage is only from their direct cast, not including the start of the next turn.
> 
> For me, the fun in this game is being handed a problem and a bunch of tools and being asked to find some clever way to use those tools. The beauty of this as a video game, as opposed to a board game, is that the computer is doing all of the work of showing you what's about to happen while you're planning. Anything that interferes with that is an instant fun-killer, IMHO. I want to be able to put together that insane clutch play, and if poison damage should affect my decisionmaking it should be reflected in the prediction mechanism.
> 
> I think I've seen other games show it in the health bar of an enemy as its own segment for "you didn't do this direct damage but the unit will be taking it". I strongly suggest doing something like that here. If poison wrecks my ability to trust predictions I will avoid those cards despite all the interesting possibilities.

The tricky thing here is that prediction shows you what will happen directly after you cast but poison only procs onTurnEnd.  but we could do something similar to suffocate (the purple bar) like you're proposing but with a green bar.

Thoughts @SoulMuncher ?

SoulMuncher:
> Thoughts @SoulMuncher ?

Ideally, prediction would always show "game-state directly after cast", and everything else would be in some generic incoming healing/damage overlay. Doing this reliably would require a huge refactor of the prediction/smart targeting code, and I would rather it not exist than be unreliable.

As an example: Smart Targeting runs all ai units at the same time, instead of having an ally turn and then an enemy turn. This means that those units turnStart/End events don't run in the correct order, which is very problematic when introducing health regen, poison, etc. Smart Targeting also doesn't use Unit.takeDamage() or Unit.die() either, so it skips onDamage/onDeath events, including modifiers like debilitate, fortify, bloat explosions, etc.

JustinWick:
> > Apologies if this is the wrong forum, but if there's `healthRegen` and `shieldRegen`, why not `manaRegen`? It's less powerful than increasing max mana (it doesn't change things that scale off max mana) so it could be balanced to be "punchy" in its intended effects without being hard to balance with unwanted synergies. It's also just a cool effect that some cards might tack on: imagine a `Prismatic Blessing` card that gives `+10 Health Regen, +10 Shield Regen, +10 Mana Regen`.
> 
> So are you envisioning that this would result in overflow mana every turn? For the players at least, since mana regen is always equal to max mana, there's actually no functional difference. The idea is that max mana shows you how much mana you'll get every turn, but overflow mana is always allowed from spells or potions and so there's no cap.

Yes, mana overlflow is both mathematically and thematically different. Mathematically it doesn't count towards "percent maximum mana" calculations, and if maximum mana is somehow reduced to zero this mana would still be granted. Thematically it's a temporary bonus that doesn't feel as bad when it's gone as reducing max mana would. Also it just feels "punchy" to see the bar overflowing and think of that sweet mana welling up, crackling and waiting to be used for mayhem!

That said, the surest way to kill any game design is to overcomplicate it. But there's such an obvious symmetry here, and the fact that mana is explicitly designed to overflow seems to "want" this feature for the reasons I give above, and probably a few more I haven't thought of--e.g. effects that trigger off of mana overflow, (now there's a thought!) associated audio/visual effects, and getting people like me to stop asking why it's not in there already :)

src/modifierManaBarrier.ts
```diff
@@ -0,0 +1,101 @@
...
+    modifier.tooltip = `${Math.floor(unit.mana * CalcMult(modifier.quantity))} ${i18n('Mana Barrier')}`
+  }
+}
+
+function CalcMult(quantity: number): number {
+  // Each Quantity = 1% barrier effectiveness
+  // 100(%) quantity = 1:1 ratio = 1 mana blocks 1 damage
+  // 20(%) quantity = 5:1 ratio = 5 mana blocks 1 damage
+  return quantity / 100;
+}
```
jdoleary, line 101: Why not 1 quantity = 10% or something.  Why out of 100?

jdoleary: This way it won't have to be upgraded so many times.  I know we discussed this the other day, but rather than making pushing the "plus button" do 5 quantity we should just change the runes to be more effective per upgrade IMO.  Thoughts?

src/modifierOverheal.ts
```diff
@@ -0,0 +1,58 @@
...
+function updateTooltip(unit: Unit.IUnit) {
+  const modifier = unit.modifiers[overhealId];
+  if (modifier) {
+    // Set tooltip:
+    modifier.tooltip = `${modifier.quantity}% ${i18n('Overheal')} ${i18n('Effectiveness')}`
+  }
+}
+
+function CalcMult(quantity: number): number {
+  // Each Quantity = 1% overheal effectiveness
```
jdoleary, line 56: Same comment as mana barrier

src/cards/soul_bind.ts
```diff
@@ -44,28 +44,33 @@ const spell: Spell = {
...
+    onTakeDamage: (unit, amount, underworld, prediction, damageDealer) => {
+      // Soul Bind splits damage/healing across all soul bound units
+      const boundUnits = getSoulBoundUnits(underworld, prediction);
+      for (let i = 0; i < boundUnits.length; i++) {
+        const boundUnit = boundUnits[i];
+        // This unit is already running the take damage event, so exclude it
+        if (boundUnit && boundUnit != unit) {
+          // Exception: Temporarily remove the bound unit's Soul Bind event to prevent infinite loop
+          const index = boundUnit.events.findIndex(e => e == soulBindId);
+          boundUnit.events.splice(index, 1);
```
jdoleary, line 56: Another way to do this is
`boundUnit.events = boundUnit.events.filter(x => x !== souldBindId)`;
I like this method because you don't have to mess with indexes.  Note you must assign it back to events on the left side of the `=`

src/modifierOnHitHealing.ts
```diff
@@ -0,0 +1,44 @@
+import { registerEvents, registerModifiers } from "./cards";
+import { getOrInitModifier } from "./cards/util";
+import { healUnit } from "./effects/heal";
+import * as Unit from './entity/Unit';
+import Underworld from './Underworld';
+
+// Restores (quantity) health per hit
+export const onHitHealingId = 'On Hit Healing';
+export default function registerOnHitHealing() {
+  registerModifiers(onHitHealingId, {
```
jdoleary, line 10: All these new modifiers need a `cost` so we can see them in the shop.  
We can deal with the variable cost later, but for now let's make them all available for testing.

src/modifierManaBarrier.ts
SoulMuncher: This is for the same reason shield/poison now give 1 shield/poison per quantity instead of 30/20. Being able to add modifiers in smaller quantities is what gives Overheal, Shield Regen, and OnHitPoison room to exist.

Debilitate may be a closer example: If we have a rune/spell/enemy that applies some amount of debilitate when it hits, we may not want to apply a full 100% increased damage every time. One source may inflict 5% while another inflicts 50%, and having 1 quantity = 1% makes sense in this case. It also lets us easily change any source without affecting another one. 

Technically, 1 quantity only ever needs to be = the lowest common denominator of each modifier: If we always apply debilitate in multiples of 10%, we _could_ make 1 quantity of debilitate = 10%, but what do we gain by doing that? Giving every modifier a different quantity-to-% ratio is confusing, and the only benefit I can see is adding modifiers through the power bar more quickly, which doesn't outweigh the balance and design limitations imo. For runes, a "quantity per upgrade" variable seems necessary either way. So I think it makes sense to have these % based modifiers all use the same 1:1% ratio. I can read an effect and know immediately that 10 quantity = 10% without having to check the modifier logic. If I want to add a 5% source, I don't have to change the base 1:10% ratio of the modifier and double the quantity applied by all other sources

src/cards/soul_bind.ts
SoulMuncher: Lol, I totally should've thought of that. I'll update this for both soul bind and thorns

src/modifierOnHitHealing.ts
SoulMuncher: True, I'll add this

src/modifierManaBarrier.ts
jdoleary: Gotcha, okay that works for me.

jdoleary (approved):

Issue #640, opened 2024-04-16
Predictions don't update player stamina bar


SoulMuncher:
Easily tested with Potion Shatter + Stamina Potion or Merge

SoulMuncher:
Stamina bar logic is in Unit.ts `syncPlayerHealthManaUI()`

## IMPORTANT
Predictions are optimized to skip some game events, including turnStart/End where stamina is regenerated, so all prediction units have stamina set = staminaMax in copyForPredictionUnit(), making a fix for the UI Stamina Bar pretty complicated.

We could remove this stamina exception and run turnStartEvents() in planningView, but without further refactors to the prediction system, it could negatively impact smart targeting/predictions in unexpected ways

SoulMuncher:
https://github.com/jdoleary/Spellmasons/issues/388

jdoleary:
Let's keep it simple and just ignore stamina predictions.

PR #869, opened 2024-07-14
Fix Player Health/Stamina/Mana bar visuals

SoulMuncher:
Fixes a bunch of issues with the player health/mana bar, and soon to fix stamina prediction issues

- The value of incoming shield is now displayed during prediction
- Changes to max Health/Mana are now shown in the lower UI bars the same way they are shown in the overhead prediction bars (i.e. blood curse shows as doubling current hp and preserving max health ratio)
- Health/Mana gain visuals were swapped such that increases give a "fuller" looking bar, and decreases give an "emptier" one
- Fixed an issue where overfilling mana would always display mana cost bars, even when no prediction was running
- Fixed inaccurate and overlapping mana cost bars
- Overfilling mana is now much more clear, and you can see each stage of mana bar as it appears, instead of mana just appearing "full" until post-prediction
- The Health/Mana bars now look the same in prediction as they will after the effect occurs (predicting blood curse no longer shows hp going to full, overfilled mana displays correctly, etc. healing/damage and mana gain/loss visuals are still present)
- Fixed "sheild" typo

## TODO
- #640

### DIFF
--- a/src/entity/Unit.ts
+++ b/src/entity/Unit.ts
@@ -53,8 +53,8 @@ import { chooseObjectWithProbability } from '../jmath/rand';
 const elCautionBox = document.querySelector('#caution-box') as HTMLElement;
 const elCautionBoxText = document.querySelector('#caution-box-text') as HTMLElement;
 const elHealthBar = document.querySelector('#health .fill') as HTMLElement;
-const elHealthBarSheild = document.querySelector('#health .fill:nth-child(2)') as HTMLElement;
-const elHealthCostSheild = document.querySelector('#health .cost:nth-child(4)') as HTMLElement;
+const elHealthBarShield = document.querySelector('#health .fill:nth-child(2)') as HTMLElement;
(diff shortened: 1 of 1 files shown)


jdoleary (approved):
Fantastic work!  I see there's still one todo left and this is still marked as a draft so I wont mere, but it look great to me.  All my tests pass

SoulMuncher:
Predictions are optimized to skip some game events, including turnStart/End where stamina is regenerated, so all prediction units have stamina set = staminaMax in copyForPredictionUnit(), making a fix for the UI Stamina Bar pretty complicated.

We could remove this stamina exception and run turnStartEvents() in planningView, but without further refactors to the prediction system, it could negatively impact smart targeting/predictions in unexpected ways, and a refactor of the prediction system is not in the scope of this PR

@jdoleary
I'm going to remove the stamina fix from the TODO, this PR is ready to go

PR #879, opened 2024-07-17
Ricochet Arrow + Arrow Improvements

### LINKED ISSUE #765: Beam spells or spells that reflect off of walls
https://discord.com/channels/1032294536640200766/1032294537235812404/1242903165129724007

@Sana

### LINKED ISSUE #854: Target arrow refunds cost of the arrow if it hits a wall, but still executes the connected spells
https://discord.com/channels/1032294536640200766/1259602595245326387/1259602595245326387

Apotheosis
OP
 — Yesterday at 4:11 PM
For instance, using a summoning and target arrow, when the arrow hits a wall it refunds the cost of the arrow saying no target found, but still summons the mob where the arrow hits (at the normal cost of the summoning spell)

### LINKED ISSUE #876: Ricochet Arrow
MΣMΣ_MΛП
OP
 — Today at 2:52 AM
Its the same as the normal target arrow except it can bounce off a wall once. Stacking the ricochet target arrow would increase the bounces. This could be a nice addition to the game and would provide strategic use

SoulMuncher:
closes #765 
closes #876 
closes #854 

Adds Target Ricochet Arrow, can bounce once per quantity
Target Arrow can pierce with quantity
Explosive arrow explodes when colliding with a wall/piercing a unit

Significant changes to arrow code:
- Now entirely physics based, all "arrow path" code removed
- Pierce and Bounce support (both as numbers, phantom arrow set to infinity pierces by default)
  - Pierce is prioritized over bounce, such that arrows will always bounce off of walls, but will only bounce off of units if unable to pierce any further.
- Some changes to projectile.ignoredUnits for collision: Ex. Piercing arrows can re-collide with units if they leave the collision area and come back
- State passed through to projectile collision function (primarily for targeting spells)
- VFX Projectiles such as the meteor can fully ignore collision and be assigned a lifetime

## TODO
- [x] Increase forcemove timeout with each bounce so the bouncing projectile doesn't timeout
see usage of config.FORCE_MOVE_PROMISE_TIMEOUT_MILLIS
- [ ] Infinite loop limit is getting triggered
    -  I'm actually going to leave this one in.  There must be some limit and 1000 is plenty.


src/Underworld.ts
```diff
@@ -466,12 +466,15 @@ export default class Underworld {
...
-          // Units will move parallel to the wall until passing it
-          // const projection = projectVelocityAlongWall(velocity, collision.wall);
-          // const newPosition = Vec.add(pushedObject, Vec.multiply(deltaTime - collision.msUntilCollision, projection));
-          // pushedObject.x = newPosition.x;
-          // pushedObject.y = newPosition.y;
+          // Stops the unit
+          //velocity.x = 0;
+          //velocity.y = 0;
+
+          // UNCOMMENT THIS WITH ANY EXAMPLES
```
jdoleary, line 473: I'm not sure what the above comment means

src/cards/arrow.ts
```diff
@@ -48,19 +48,14 @@ const spell: Spell = {
...
 };
-export function arrowEffect(multiShotCount: number, collideFnKey: string, doesPierce: boolean = false) {
+export function arrowEffect(multiShotCount: number, collideFnKey: string, piercesRemaining: number = 0, bouncesRemaining: number = 0) {
   return async (state: EffectState, card: ICard, quantity: number, underworld: Underworld, prediction: boolean, outOfRange?: boolean) => {
     let targets: Vec2[] = state.targetedUnits;
-    const path = findArrowPath(state.casterPositionAtTimeOfCast, state.castLocation, underworld)
-    targets = targets.length ? targets.map(t => {
-      const path = findArrowPath(state.casterPositionAtTimeOfCast, t, underworld);
-      return path ? path.p2 : state.castLocation;
-    }) : [path ? path.p2 : state.castLocation];
```
jdoleary, line 58: Was this already unused? I don't see any reference to it

src/cards/meteor.ts
```diff
@@ -133,11 +139,13 @@ async function meteorProjectiles(meteorLocations: Vec2[], underworld: Underworld
     makeForceMoveProjectile({
       pushedObject,
       startPoint: startPos,
-      endPoint: meteor.destination,
-      speed: distanceOffset / meteor.travelTime,
-      doesPierce: true,
-      ignoreUnitIds: underworld.units.map(u => u.id),
+      velocity,
+      piercesRemaining: 10000,
```
jdoleary, line 143: Use `Infinity` which is an actual number in js.  That way subtracting from it will not decrease it

src/cards/phantom_arrow.ts
```diff
@@ -22,7 +22,8 @@ const spell: Spell = {
     animationPath: '',
     sfx: 'phantomArrow',
     description: ['spell_phantom_arrow', damageDone.toString()],
-    effect: arrowEffect(1, phantomArrowCardId, true)
+    // Phantom arrow has "infinte" pierce built in
+    effect: arrowEffect(1, phantomArrowCardId, 10000, 0)
```
jdoleary, line 26: `Infinity` instead of 10000`

src/cards/target_arrow.ts
```diff
@@ -33,94 +34,91 @@ const spell: Spell = {
...
+  events: {
+    onProjectileCollision: ({ unit, pickup, underworld, projectile, prediction }) => {
+      if (projectile.state) {
+        if (unit) {
+          addTarget(unit, projectile.state, underworld, prediction);
+        } else if (pickup) {
+          addTarget(pickup, projectile.state, underworld, prediction);
+        } else {
+          // Add collision location as target
+          //addTarget(projectile.pushedObject, projectile.state, underworld, prediction);
```
jdoleary, line 116: I think prev functionality allowed for spawning a unit at the wall where the target arrow collided

src/cards/target_ricochet_arrow.ts
```diff
@@ -0,0 +1,127 @@
...
+      // Update targets variable to include newly added targets
+      targets = getCurrentTargets(state);
+      if (!prediction && !globalThis.headless && globalThis.predictionGraphics) {
+        const promises: Promise<void>[] = [];
+        targets.forEach(t => {
+          // Animations do not occur on headless
+          promises.push(new Promise<void>((resolve) => {
+            if (globalThis.predictionGraphics) {
+              globalThis.predictionGraphics.lineStyle(2, colors.targetingSpellGreen, 1.0)
+              playSFXKey('targetAquired');
```
jdoleary, line 97: When animating, the target ricochet arrow doesn't show targets being acquired as it collides with them.  I think this would look/feel better

jdoleary: I think maybe I just need to make this an automatic part of addTarget so we don't have to handle it individually in each targeting spell so actually leave that out of this for now and I'll address later

jdoleary (commented):
So cool!! Lemme know when ready for final review!

src/cards/arrow.ts
SoulMuncher: > @jdoleary
> Was this already unused?

In the main `arrowEffect` function, `findArrowPath()` was used to pre-determine an "endpoint" for the arrow, so that forcemoves could remove it when it hit a wall, but now that's done via physics collision. `findArrowPath()` was also responsible for setting Target Arrow's cast location and preventing weird interactions like placing a Target Circle out of range/on the other side of a wall, but this is also handled through collision now

SoulMuncher:
@jdoleary This should be good to go now

jdoleary:
Outstanding todo
- [ ] Increase forcemove timeout with each bounce so the bouncing projectile doesn't timeout
see usage of config.FORCE_MOVE_PROMISE_TIMEOUT_MILLIS

jdoleary (approved):

PR #894, opened 2024-07-25
Improved targeting prediction

### LINKED ISSUE #721: runPrediction() on MouseMove is different from realtime result
Hard to explain, but I'll get a video.

If a unit is on the edge of a targeting radius, such as target cone, and you move the mouse one pixel away so that the unit is no longer in the radius, it will still PREDICT targeting the enemy, but the actual spell will not.

This is problematic because it's common to try and "perfectly line up" a targeting spell to hit a certain number of enemies, and all that effort will be wasted on a spell not doing what it says it will.

### LINKED ISSUE #893: Incorrect Prediction: Cast Location
Happens more frequently on Jamison's computer.  Seems to be that cast location changes slightly between last prediction and click.
Affects many bounce + arrow or cone + slash for example.

Possibly fixable by saving last cast location from last completed prediction and overwriting cast location on click with it if it's close enough in space and time.

Discovered: The "prediction" issue is NOT a force move issue, it happens when the last prediction is different from the target sent in the spell.

jdoleary:
closes #893 
closes #721 

Attempt to solve the "target cone + slash" and the "many add_bounce + arrow" issue where the actual spell can cast with slightly different coordinates than the last successful runPrediction which leads to the frustrating scenario where the prediction appears to have been wrong, but the game just sent the cast with very slightly different coordinates than the last calculated prediction.

### DIFF
--- a/src/graphics/PlanningView.ts
+++ b/src/graphics/PlanningView.ts
@@ -547,6 +547,10 @@ export async function runPredictions(underworld: Underworld) {
   }
   const startTime = Date.now();
   const mousePos = underworld.getMousePos();
+  // Queue mousePosition from the start of runPredictions so that when
+  // runPredictions is successful it can be saved to
+  // check if castLocation is different from last successful runPredictions
+  globalThis._queueLastPredictionMousePos = mousePos;
   // Clear the spelleffectprojection in preparation for showing the current ones
   clearSpellEffectProjection(underworld);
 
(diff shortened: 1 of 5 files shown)


SoulMuncher:
Looks good to me! I tested with target column, and in places where I would expect the bug to happen it's instead sending the "Quality of Life: Overriding mouse position with last successful runPrediction mouse position." log and prediction is accurate.

Issue #182, opened 2023-12-17
Mana cost at start of spell


SoulMuncher:
Cloning yourself creates a spellmason with full mana (i.e.) not including the mana cost of clone. Clone effect happens before mana is taken. May affect other spells too, such as split.

jdoleary:
I'm okay with leaving this one as is.  As long as it clones the mana you had before casting clone.  Mana is taken after a spell is cast.  I don't want to bring in a special exception for clone.
I would expect if you have 180/200 mana and clone yourself that the clones would have 180/200 mana too.
I just verified that this is indeed what happens so I'm going to close this.

We need to avoid exceptions in code logic at all costs and only use as a total last resort - it drastically increases the maintainability burden.

SoulMuncher:
Re-opening this:

Split self "Insufficient mana" issue: because mana is taken after split instead of before. Predicted mana after split is really inaccurate as well, possibly due to this.

In most games, a spell takes mana before the effect occurs/completes, instead of after. Instead of this being an edge case for split and clone, I think we should move mana cost to the start of spells.

jdoleary:
This may have some unexpected consequences, but I'm open to try it.

SoulMuncher:
Would help fix the clone/merge combo too

jdoleary:
TODO: Reenable mana barrier once this is complete and test that it fixes #947

jdoleary:
Okay so the reason it works this way is so that spells that _give_ you mana allow you to cast more.  For example mana steal.  If you do a bunch of mana steal and want to keep your targets and then build a bigger spell off of that, this allows you to do so.
I'm not sure we can have the best of both worlds (being able to split without enough mana after the spell to cast it AND being able to use mana from mana steal in the same spell). Thoughts @SoulMuncher ?

SoulMuncher:
@jdoleary 
As implemented now: it spends mana after each individual spell right?

I think we can get away with spending the mana in steps still, we just need to put the costs before each spell effect instead of after. Mana Steal should still work, mana barrier/split should be fixed, and mana scamming should be much harder to do.

[Circle + Mana Steal + Slash]
- spend mana for circle
- circle
- spend health for mana steal
- mana steal
- spend newly aquired mana for slash
- slash

May also need to update refund logic to ensure it doesn't create some kind of mana duplication exploit.

PR #918, opened 2024-07-29
Runes 2

SoulMuncher:
### Changes:
- Fix onDealDamage events not updating damage amount
- adds OnKill modifier event
- adds OnPickup modifier event
- adds Bounty System (Active when you have relevant runes)

### Runes:
- Potion Effectiveness
- Potion Barrier - Grant shield on pickup potion
- Shield Bash - Consume shield to deal extra damage
- On Bounty Kill: Restore Health/Stamina/Mana
- Extra damage against units with a bounty
- Restore Mana on kill
- Liquidmancer - Units take % more damage from liquid
- Heavy Impact - Units take % more impact damage
  - Technically increases the magnitude of impact, which both increases impact damage and reduces heavy impact threshold
- Affinity Runes - Category spells are % cheaper
  - Witch no longer reduces the cost of curses, just grants a cheaper/empowered contaminate
- Inexhaustbile - Movement spells have no expense scaling

### Other:
I want to add a rune that would reset expense scaling on death/on bounty kill, but noticed that cardUsageCounts is handled after castCards resolves, which makes these effects function unexpectedly. Related: #182 

## TODO
- Add more runes


src/Underworld.ts
```diff
@@ -2126,9 +2136,22 @@ export default class Underworld {
...
     for (let e of enemies) {
       this.spawnEnemy(e.id, e.coord, e.isMiniboss);
     }
+    // if any players have the bounty hunter modifier, add a bounty to a random unit's head at the start of level
+    // to get a bounty, a unit must be alive, in the enemy faction, not a doodad, and not yet have a bounty
+    if (this.players.some(p => p.unit.modifiers[bountyHunterId])) {
+      let units = this.units;
+      units = units.filter(u => u.alive && (u.faction == Faction.ENEMY) && (u.unitSubType != UnitSubType.DOODAD) && !u.modifiers[bountyId]);
+      if (units.length > 0) {
+        const chosenUnit = units[randInt(0, units.length - 1)];
```
jdoleary, line 2149: This should be seeded so that each client will get the same
You _could_ use randInt with a seed but I'd recommend `chooseOneOfSeeded` instead.

Seeds can be make like so:
`const seed = seedrandom(`STRING HERE`);`
Where the string is used to ensure the seed remains the same given certain circumstances.  For example if we want the random generator to return something unique for each game and each player we add those to the seed string (see `getUniqueSeedString` but in this case I think it only has to be unique to the underworld seed and underworld levelIndex)

src/cards/util.ts
```diff
@@ -6,6 +6,7 @@ export interface Modifier {
     quantity: number;
     // If true, the modifier will not be removed from the unit on death 
     keepOnDeath?: boolean;
+    keepBetweenLevels?: boolean;
```
jdoleary, line 9: removeModifier already prevents removing runes.  Is `keepBetweenLevels` necessary?
```  
if (modifier && modifier.costPerUpgrade) {
    // Modifier is a Rune and should NOT be removed
    return;
  }```

File: src/entity/Unit.ts [hunk 3]
```diff
@@ -981,6 +984,23 @@ export function takeDamage(damageArgs: damageArgs, underworld: Underworld, predi
...
+            fn(sourceUnit, unit, underworld, prediction);
+          }
+        }
+      }
+    }
+
+    // Special case: Bounty modifier persists on death and is removed AFTER onKill events run
+    if (unit.modifiers[bountyId]) {
+      removeModifier(unit, bountyId, underworld);
+    }
```
jdoleary, line 1003: Rather than have a special case here and need to add keepOnDeath again, let's just add `sourceUnit` to `die` and run the onKill event stuff inside of `die`.
What do you think?

src/cards/util.ts
jdoleary: I think I know why you added this back, see my comment in Unit.ts for more rationale for removing it

src/entity/Unit.ts
```diff
@@ -1457,6 +1477,17 @@ export async function runTurnEndEvents(unit: IUnit, underworld: Underworld, pred
   ));
 }
 
+export async function runPickupEvents(unit: IUnit, pickup: IPickup, underworld: Underworld, prediction: boolean) {
+  await Promise.all(unit.events.map(
```
jdoleary, line 1481: This should be covered by `raceTimeout` just in case an event never resolves

src/modifierAffinityDamage.ts
```diff
@@ -0,0 +1,20 @@
...
+import { getOrInitModifier } from "./cards/util";
+import * as Unit from './entity/Unit';
+import Underworld from './Underworld';
+
+export const affinityDamageId = 'Affinity: Damage';
+
+export default function registerDamageAffinity() {
+  registerModifiers(affinityDamageId, {
+    description: i18n('Damage spells cost [quantity]% less mana'),
+    costPerUpgrade: 40,
```
jdoleary, line 11: IMO, Affinity runes should cost more than 40.  A cost of 40 means they could max it out at a 50% reduction after only 2 turns (200 skill points).  That seems OP.  
This is what I would recommend:
Blessing: 40
Curse: 50
Movement: 40
Soul: 70
Targeting: 80
Damage: 80

Thoughts?

src/modifierBounty.ts
```diff
@@ -0,0 +1,56 @@
...
+import floatingText from "./graphics/FloatingText";
+import { bountyHunterId } from "./modifierBountyHunter";
+import Underworld from './Underworld';
+import { skyBeam } from "./VisualEffects";
+
+export const bountyId = 'Bounty';
+export const bountyColor = 0xffdc64;
+export default function registerBounty() {
+  registerModifiers(bountyId, {
+    description: 'This unit has a bounty. Killing them will trigger bounty effects.',
```
jdoleary, line 14: SUCH A COOL IDEA :D

src/modifierPotionEffectiveness.ts
```diff
@@ -0,0 +1,36 @@
...
+      });
+    }
+  });
+  registerEvents(potionEffectivenessId, {
+    onPickup: async (unit: Unit.IUnit, pickup: Pickup.IPickup, underworld: Underworld, prediction: boolean) => {
+      const modifier = unit.modifiers[potionEffectivenessId]
+      if (modifier) {
+        if (pickup.name.includes("Potion")) {
+          // Pickup.power determines the effectiveness of the potion, so we multiply it
+          pickup.power *= CalcMult(modifier.quantity);
```
jdoleary, line 26: Note to self: Test with prediction

src/Underworld.ts
```diff
@@ -427,28 +430,35 @@ export default class Underworld {
...
-        // console.log("Impact Damage: ", impactDamage, prediction);
+        // Heavy Impact if magnitude > 2
+        if (magnitudeOfImpact > 2) {
+          // Calculate impact damage using magnitude
+          const impactDamage = Math.floor((magnitudeOfImpact - 2) * 10);
 
-        // If impact damage > 0, we hit the wall hard enough for
-        // a "heavy impact", which stops the object and damages units
-        // otherwise, make the unit slide along the wall
-        if (impactDamage > 0) {
```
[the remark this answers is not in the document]
SoulMuncher: It was changed to the code below. It's functionally the same right now, but it reads better and leaves more room for other 'heavy impact' events/runes in the future (it's a cool mechanic and I think we could do some fun things with it, like triggering an explode or something). Code wise: Velocity/Magnitude of impact is the condition that needs to be met for a heavy impact, and impact damage is one result of the event

```typescript
// Heavy Impact if magnitude > 2
if (magnitudeOfImpact > 2) {
    // Calculate impact damage using magnitude
    const impactDamage = Math.floor((magnitudeOfImpact - 2) * 10);```

src/cards/util.ts
SoulMuncher: Here was the thought process:
- Want to add a Bounty System: One random unit is given a bounty, and killing that unit grants special rewards/effects. It creates a new assassin-like playstyle with unique challenges, strategies, etc.
- Players can engage with the bounty system via runes that have bounty related effects, such as restoring health/stamina/mana when they claim a bounty
- Bounties should only begin appearing if the player has at least one of these runes, as bounties have no innate purpose/effect without them
- The system needs some kind of handler for spawning bounties. Putting it in each individual modifier is not maintainable, and putting it in Underworld turn logic feels invasive/incorrect
- Decided to write this system through a modifier called Bounty Hunter:
  - The Bounty Hunter modifier is what controls the bounty system by spawning bounties and highlighting them through onDrawSelected and stuff like that
  - Any bounty related rune will also give the player a persistent Bounty Hunter modifier to effectively "enable" the bounty system
  - Bounty Hunter needs to be persistent through levels, like the runes it comes with, but it should not share any other features of runes (has no effect on its own, isn't purchasable)

src/entity/Unit.ts
SoulMuncher: Oh, running the onKill events inside of die()  is _way_ better. I'll fix that.

src/modifierAffinityDamage.ts
SoulMuncher: Hmm... Yeah, 40 is a little low. Here's what I'm thinking:

Generally speaking, 200 points into an affinity should be better for casting spells of that category/playstyle than 200 points into mana, and they should provide roughly equal power when part of their best synergies.

Things to consider
- Max mana
  - Helps with all categories of spells
  - Noncommittal: You can buy as much or as little as wanted, doesn't lock you into one playstyle/category of spells, easier to take early on
  - Scales with effects that rely on max mana
  - Scales well with rune effects like Mana Barrier
- Affinity 
  - Much larger effect on the specific category of spell
  - Committal: Best value at max purchase (200 points), locks you into playing around that category of spells
  - Does not scale with effects that rely on max mana
  - Scales better with flat mana recovery like from potions or mana steal
 
The balance of affinity runes vs mana is **100%** dependent on what other effects get added into the game, and every mana-related effect we add will make on or the other more valuable than before.

As far as specific values go, I like the idea of all affinities costing the same amount, and that would be balanced by what other runes are available along that playstyle. I'd also want the costs to stay semi-low. 80 feels too high because you spend half the game getting the reduced mana cost, and that doesn't leave much time for combining other "cooler" rune effects. Especially since affinities "all or nothing" compared to other runes, which can be bought once to enable some kind of combo.

I could bump them all to 60 and we balance from there? Oh, and I did leave mana affinity out. I don't really want to, because I love the mana manipulation spells and playstyle, but I'm not sure how to approach it because any changes or added rune effects seem prone to enabling infinite combos

src/cards/util.ts
jdoleary: Hmm, I see, okay I'm on board

jdoleary: Do you forsee this changing over the lifetime of the game for a modifier instance, because if not I think we should put it in the modifier source rather than the modifier instance state. This will allow us to just check for it in `removeModifier` (in the same place that we check `costPerUpgrade`

src/modifierAffinityDamage.ts
jdoleary: Hmm okay good points.  
I kind of agree, but the damage category feels different to me.  I just think getting to 50% on damage within the first two levels for instance feels super OP.  

The mana upgrade only gives you 5 mana and costs 30; so it would take 12 clicks to double from 60 to 120 (3.6 levels cleared).

Let's try having the affinities cost 60 and go from there (tho I do imagine no one's gunna pick a Blessing Affinity for that cost cause it's just not as powerful as a damage affinity

SoulMuncher: That's true... I'm hoping all of the armor/thorns/similar effects are going to make blessings competitive. In this PR I'd actually argue that MOVEMENT is the strongest category. Liquidmancer / Heavy Impact with all the extra mana/cost reduction you can get is insane.

The soul category is an important one to keep an eye on too, with all the weird stuff it can do. And summons.

src/cards/util.ts
SoulMuncher: I'm not 100% sure what you mean by forsee it changing. I don't think bounty hunter would ever _not_ persist through death, if that's what you're asking. I'll move it to the `Modifiers` interface with `costPerUpgrade`

jdoleary (approved):
Excellent work!

PR #921, opened 2024-07-30
Fix rune choosing

jdoleary:


jdoleary:
TODO: Unlocking a rune can still make it appear somewhere else

PR #952, opened 2024-08-09
Runes 3

SoulMuncher:
### Improvements:
- Finished setting up [modifier staging](https://github.com/jdoleary/Spellmasons/pull/937) to ensure modifiers/runes run in the expected order
- Replaced old updateTooltip() functions with the new [OnTooltip()](https://discord.com/channels/@me/1179829034100740136/1268300680666415206) event for all modifiers
- [Removed intermittent rounding](https://discord.com/channels/@me/1179829034100740136/1268942431626461317) from many spells/modifiers such as split, fortify, and mana barrier. Also updated UI to parse decimals
- [Balance](https://discord.com/channels/@me/1179829034100740136/1269812842002124874)
  - Stat Upgrades: Cost 30 per upgrade (from 25)
  - Mana: +5 per upgrade (from 10)
  - Armor: Removed upgrade limit (from 5)
  - Armor: Cost 20 per upgrade (from 30)
  - Dare Devil: Half health is now a persistent multiplier similar to Far Gazer's reduced stamina
  
### New Runes:
- WIP
  
## TODO
  - Evaluate [staging for the shield modifier](https://discord.com/channels/1032294536640200766/1266746593952927795/1271249061646831667) while considering interactions with thorns, mana barrier, etc.
  - New Runes

### DIFF
--- a/src/cards/blood_curse.ts
+++ b/src/cards/blood_curse.ts
@@ -30,8 +30,8 @@ function add(unit: IUnit, underworld: Underworld, prediction: boolean) {
     Unit.addEvent(unit, id);
     // This is done in first time setup because the modified stats are stored on the unit
     // otherwise it would apply the multiplier each time the unit is loaded
-    unit.healthMax = Math.floor(unit.healthMax * healthMultiplier);
-    unit.health = Math.floor(unit.health * healthMultiplier);
+    unit.healthMax *= healthMultiplier;
+    unit.health *= healthMultiplier;
   });
 
   if (unit.modifiers[suffocateCardId]) {
@@ -41,8 +41,8 @@ function add(unit: IUnit, underworld: Underworld, prediction: boolean) {
 
 function remove(unit: IUnit, underworld: Underworld) {
   // Health should not go below 1
-  unit.healthMax = Math.max(1, Math.floor(unit.healthMax / healthMultiplier));
-  unit.health = Math.max(1, Math.floor(unit.health / healthMultiplier));
+  unit.healthMax = Math.max(1, unit.healthMax / healthMultiplier);
+  unit.health = Math.max(1, unit.health / healthMultiplier);
 }
 
 const imageName = 'spellIconBloodCurse.png';
--- a/src/cards/blood_curse.ts
+++ b/src/cards/blood_curse.ts
@@ -34,10 +34,6 @@ function add(unit: IUnit, underworld: Underworld, prediction: boolean) {
     unit.health = Math.floor(unit.health * healthMultiplier);
   });
 
-  if (!prediction) {
-    updateTooltip(unit);
-  }
-
   if (unit.modifiers[suffocateCardId]) {
     updateSuffocate(unit, underworld, prediction);
   }
@@ -47,14 +43,6 @@ function remove(unit: IUnit, underworld: Underworld) {
   // Health should not go below 1
   unit.healthMax = Math.max(1, Math.floor(unit.healthMax / healthMultiplier));
   unit.health = Math.max(1, Math.floor(unit.health / healthMultiplier));
-}
-
-
-export function updateTooltip(unit: Unit.IUnit) {
-  if (unit.modifiers[id]) {
-    // Set tooltip:
-    unit.modifiers[id].tooltip = `Blood Curse`;
-  }
 }
 
 const imageName = 'spellIconBloodCurse.png';
@@ -82,6 +70,13 @@ const spell: Spell = {
     remove,
   },
   events: {
+    onTooltip: (unit: Unit.IUnit, underworld: Underworld) => {
+      const modifier = unit.modifiers[id];
+      if (modifier) {
+        // Set tooltip:
+        modifier.tooltip = `Blood Curse`;
+      }
+    },
     onTakeDamage: (unit: IUnit, amount: number, _underworld: Underworld, prediction: boolean, damageDealer?: IUnit) => {
       // Takes healing as damage
       if (amount < 0) {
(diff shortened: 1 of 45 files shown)


src/cards/blood_curse.ts
```diff
@@ -82,6 +70,13 @@ const spell: Spell = {
     remove,
   },
   events: {
+    onTooltip: (unit: Unit.IUnit, underworld: Underworld) => {
+      const modifier = unit.modifiers[id];
+      if (modifier) {
+        // Set tooltip:
+        modifier.tooltip = `Blood Curse`;
+      }
+    },
```
jdoleary, line 79: onTooltip is meant to run every gameLoop.  If the modifier is just the same as the id then it doesn't need to be set here because the id will be used instead.

src/cards/fortify.ts
```diff
@@ -112,4 +112,11 @@ function add(unit: Unit.IUnit, _underworld: Underworld, _prediction: boolean, qu
     }
   });
 }
-export default spell;
+
+function CalcMult(quantity: number): number {
+  // Each Quantity cuts incoming damage in half multiplicatively
+  const damageReduction = Math.pow(DAMGAGE_REDUCTION_PROPORTION, quantity);
```
jdoleary, line 118: Should we balance this to do 2,4,6,8 instead of 2,4,8,16?

src/cards/index.ts
```diff
@@ -779,9 +779,14 @@ export function getMaxRuneQuantity(modifier: Modifiers) {
   return Infinity
 }
 
-export type MODIFIER_STAGE = 'Soul Bind' | 'Amount Multiplier' | 'Amount Flat' | 'Amount Override'
+// Modifiers should not be in the same stage if 
+// swapping their order changes the outcome of the event.
+// I.E. Debilitate and Armor can't be in the same stage,
+// but Debilitate and Fortify are mostly interchangeable
```
jdoleary, line 785: Excellent context! Thank you!

src/cards/poison.ts
```diff
@@ -39,7 +39,6 @@ const spell: Spell = {
     },
   },
   modifiers: {
-    stage: 'Amount Flat',
```
jdoleary, line 42: No stage for poison?

src/cards/split.ts
```diff
@@ -25,17 +25,16 @@ function remove(unit: Unit.IUnit, underworld: Underworld) {
...
+  const multiplier = Math.pow(inverseMult, modifier.quantity);
+  unit.healthMax = Math.max(1, unit.healthMax * multiplier);
+  unit.health = Math.max(1, unit.health * multiplier);
+  // Note: manaMax is not changed or else split would render casters useless
+  unit.mana = unit.mana * multiplier;
+  unit.manaPerTurn = unit.manaPerTurn * multiplier;
+  unit.staminaMax = unit.staminaMax * multiplier;
+  unit.stamina = unit.stamina * multiplier;
+  unit.damage = unit.damage * multiplier;
+  unit.moveSpeed *= multiplier;
```
jdoleary, line 37: This probably needs some manual testing, I remember we had issues with split in the past when it was added and removed

src/entity/Unit.ts
```diff
@@ -1020,17 +1018,22 @@ export function syncPlayerHealthManaUI(underworld: Underworld) {
   }
   const predictionPlayerUnit = underworld.unitsPrediction.find(u => u.id == globalThis.player?.unit.id);
 
+  // Turns decimals into UI friendly numbers
```
jdoleary, line 1021: Can you add a comment about that we need to use ceil so 0.3 health doesn't round down to 0

src/modifierStatUpgrades.ts
```diff
@@ -61,7 +62,10 @@ export default function registerStatUpgradeModifiers() {
 function modifyStatBumpAmount(statBump: number, unitStatKey: "attackRange" | "manaMax" | "healthMax" | "staminaMax", player: IPlayer): number {
   switch (unitStatKey) {
     case "healthMax": {
-      // No current modifiers
+      // Dare devil gets half hp per quantity
+      if (player.unit.modifiers[dareDevilId]) {
+        statBump *= Math.pow(0.5, player.unit.modifiers[dareDevilId].quantity);
+      }
```
jdoleary, line 68: oooh yes! Good balance!

src/cards/split.ts
SoulMuncher: I think the major issue was rounding errors:

Golem hp during split: 30 -> 15 -> 7 > 3
Golem hp during purify: 3 -> 6 -> 12 -> 24

We had to use a for loop to simulate each step of "rounding" applied by each quantity of split, but since it doesn't round between steps anymore, we can just get the total multiplier and apply it immediately. I did a quick test on a couple of archers, and they seem to split/purify as expected.

I didn't do thorough testing with other modifiers involved, though

src/cards/poison.ts
SoulMuncher: I don't think so. There aren't any other onTurnEnd modifiers it can't overlap with in the unstaged group. Amount X is more for modifying a return value anyway (armor, damage/healing effectiveness, potion power, etc.)

src/cards/fortify.ts
SoulMuncher: I don't think so. It only lasts one turn and the expense scaling is super high. And intuitively + based off the description I think it would be multiplicative

src/cards/blood_curse.ts
SoulMuncher: I think this was done a long time ago to prevent it from saying "Blood Curse 1"
It can be removed now?

jdoleary: Oh let's leave it then

jdoleary (approved):

Issue #938, opened 2024-08-05
Balance: Slow


jdoleary:
https://discord.com/channels/1032294536640200766/1269935681808109680/1269935681808109680

Issue #930, opened 2024-08-01
Bug: Timemason rune lets you upgrade multiple times but


jdoleary:
the loss doesn't take quantity into account and further mana upgrades doesn't take quantity into account.

PR #957, opened 2024-08-10
Runes 4

SoulMuncher:
### Improvements
- Cleanup bounty system
- Fixed random selection for bounty
- Slow can now stack multiplicatively with any incoming quantity (not restricted to 20% per stack)
- Doodads cannot trigger onKill
- Slow costs 15 (from 20) and slows by 30% (from 20%) #938 
- Fix Timemason Quantity #930 

### Runes
- Good Looking - Convert [quantity] non-miniboss enemies to ally faction on spawn
- Plague Bringer - Inflict [quantity] poison to random enemies each turn
- Creeping Death - Inflicts a random enemy with [quantity] stacks of suffocate on kill (can chain)
- OnKillResurrect - [quantity]% chance to resurrect an enemy to your side on kill
- Golemancer - Spawn up to [quantity] ally golems each turn
- Alchemist - [quantity]% chance to summon a potion each turn
- Secret Ingredients - Empower the nearest potion each turn (potion power increased by [quantity])
- Heavy Toxins - Inflicting poison also inflicts a stacking slow (more slow with quantity)
- Bounty: Potion
- Bounty: Portal
- Bounty: Golem
- Bounty: Suffocate
- Bounty: Refresh - Places a new bounty immediately when one is claimed
- Rune Changes
  - Bounty: Stamina sets stamina to 100% stamina instead of giving flat stamina
  - Removed Bounty: Restore Health and Bounty: Restore Mana (too many bounty runes, these are somewhat unoriginal and overlap with portal)
  - Potion Effectiveness: Cost 30 (from 60), grants 10% effectiveness (from 50%)
  - Maybe some other balance

## TODO
- QA
- Balance
- More golemancer themed runes
  - Summoned golems are replaced with blood golems ~200sp
  - Summoned golems gain shield equal to yours
  - Something with sacrifice?


src/cards/slow.ts
```diff
@@ -50,20 +59,23 @@ const spell: Spell = {
...
       // .filter: only target living units
       const targets = state.targetedUnits.filter(u => u.alive);
       if (targets.length) {
         await Promise.all([playDefaultSpellAnimation(card, targets, prediction), playDefaultSpellSFX(card, prediction)]);
         for (let unit of targets) {
-          Unit.addModifier(unit, slowCardId, underworld, prediction, quantity);
+          // Each application of slow is multiplicative, so we must add the modifier like this
+          for (let i = 0; i < quantity; i++) {
+            Unit.addModifier(unit, slowCardId, underworld, prediction, slowPercentage);
+          }
```
jdoleary, line 78: Why not just pass the quantity to addModifier and do the loop inside the add function?
Would that interfere with other applications of slow like in poison or could moving the loop to the slow's add function mean we don't need both loops elsewhere?

src/modifierAlchemist.ts
```diff
@@ -0,0 +1,82 @@
...
+    }
+  }
+
+  if (randomCoord == undefined) {
+    console.warn('Could not find valid spawn point for golem, returning summoner position');
+    randomCoord = { x: unit.x, y: unit.y }
+  }
+
+  return randomCoord;
+}
```
jdoleary, line 82: Let's just use underworld.findValidSpawn instead of making a new function.

src/modifierBounty.ts
```diff
@@ -53,4 +56,31 @@ export default function registerBounty() {
...
+  let units = prediction ? underworld.unitsPrediction : underworld.units;
+
+  // Get existing bounty targets
+  const unitsWithBounty = getUnitsWithBounty(underworld, prediction);
+  // Max bounties = number of bounty hunters on team
+  const maxBounties = units.filter(u => u.faction == bountyHunter.faction && u.modifiers[bountyHunterId]).length;
+  if (ignoreMax || unitsWithBounty.length < maxBounties) {
+    // Find a random enemy unit and give it a bounty
+    // Unit must be alive, in enemy faction, not a doodad, and not yet have a bounty
+    units = units.filter(u => u.alive && (u.faction != bountyHunter.faction) && (u.unitSubType != UnitSubType.DOODAD) && !u.modifiers[bountyId]);
```
jdoleary, line 71: Filter out `u.flaggedForRemoval` too.

This property is true when the unit is prepared to be removed.

jdoleary: `alive` _probably_ covers this case anyway but it's probably safer to check for it despite

jdoleary: Actually yeah we should definitely check for it.  It'll cover cases like when a unit is merged.  It never dies but one of them get's flaggedForRemoval and eventually removed buy may still exist transiently

src/modifierBountyRefresh.ts
```diff
@@ -0,0 +1,34 @@
...
+    }
+  });
+  registerEvents(bountyRefreshId, {
+    onKill: async (unit: Unit.IUnit, killedUnit: Unit.IUnit, underworld: Underworld, prediction: boolean) => {
+      const modifier = unit.modifiers[bountyRefreshId];
+      if (modifier) {
+        if (killedUnit.modifiers[bountyId]) {
+          // This ignores the max bounty limit because the bounty modifier
+          // cannot be removed from the killed unit yet
+          placeRandomBounty(unit, underworld, prediction, true);
```
jdoleary, line 29: Could this run into a weird race condition if at some point there are more bounties than there should be?
Maybe the placeRandomBounty function could just ignore counting bounties on dead units because we know they are about to be processed and removed so we don't need an exception here?

src/modifierCreepingDeath.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+  });
+  registerEvents(creepingDeathId, {
+    onKill: async (unit: Unit.IUnit, killedUnit: Unit.IUnit, underworld: Underworld, prediction: boolean) => {
+      const modifier = unit.modifiers[creepingDeathId];
+      if (modifier) {
+        if (killedUnit) {
+          let units = prediction ? underworld.unitsPrediction : underworld.units;
+          // Find a random enemy unit and suffocate it
+          // Unit must be alive, in enemy faction, not a doodad, and not already suffocating
```
jdoleary, line 32: Why not allow it to add to the stacks of already suffocating units?

src/modifierGolemancer.ts
```diff
@@ -0,0 +1,124 @@
...
+import seedrandom from "seedrandom";
+import { Vec2 } from "./jmath/Vec";
+import { prng, randFloat } from "./jmath/rand";
+import { COLLISION_MESH_RADIUS } from "./config";
+import { allUnits } from "./entity/units";
+import { UnitType } from "./types/commonTypes";
+import { makeRisingParticles } from "./graphics/ParticleCollection";
+import { BLOOD_GOLEM_ID } from "./entity/units/bloodGolem";
+
+// Summon up to [quantity] golems each turn
```
jdoleary, line 16: "and grants Summon Golem card to player"

src/modifierGolemancer.ts
```diff
@@ -0,0 +1,124 @@
...
+      if (modifier && unit.alive) {
+        spawnGolems(unit, modifier.quantity, underworld, prediction);
+      }
+    },
+    onSpawn: (unit: Unit.IUnit, underworld: Underworld, prediction: boolean) => {
+      const modifier = unit.modifiers[golemancerId];
+      if (modifier && unit.alive) {
+        spawnGolems(unit, modifier.quantity, underworld, prediction);
+      }
+    }
```
jdoleary, line 51: Maybe we should just trigger turn start modifiers on spawn already so we don't have to handle everything twice? (especially for future runes)?

src/modifierGolemancer.ts
```diff
@@ -0,0 +1,124 @@
...
+  });
+}
+
+function spawnGolems(unit: Unit.IUnit, quantity: number, underworld: Underworld, prediction: boolean) {
+  let allyGolems = getLivingAllyGolems(unit, underworld, prediction);
```
jdoleary, line 56: I think the description is confusing me, so it only spawns up to quantity of golems but won't go beyond?

src/modifierGolemancer.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+export function getLivingAllyGolems(unit: Unit.IUnit, underworld: Underworld, prediction: boolean) {
+  let units = prediction ? underworld.unitsPrediction : underworld.units;
+  units = units.filter(u => u.alive && u.faction == unit.faction && [golem_unit_id, BLOOD_GOLEM_ID].includes(u.unitSourceId));
+  return units;
+}
+
+export function findRandomSummonLocation(unit: Unit.IUnit, radius: number, underworld: Underworld, prediction: boolean, seed: prng): Vec2 | undefined {
```
jdoleary, line 99: Let's import a similar function rather than making a new one.  If you want something close to the player use underworld.findValidSpawn and if you want something farther import from Summoner or Deathmason

src/modifierGoodLooking.ts
```diff
@@ -0,0 +1,46 @@
+import { registerEvents, registerModifiers } from "./cards";
+import { getOrInitModifier } from "./cards/util";
+import * as Unit from './entity/Unit';
+import * as config from './config';
+import Underworld from './Underworld';
+import { chooseOneOfSeeded, getUniqueSeedString, randInt } from "./jmath/rand";
+import { UnitSubType } from "./types/commonTypes";
+import seedrandom from "seedrandom";
+
+// Converts [quantity] random non-miniboss units to the player's faction on spawn
```
jdoleary, line 10: SO COOL!!!

src/modifierGoodLooking.ts
```diff
@@ -0,0 +1,46 @@
...
+function convertRandomUnitToMyFaction(unit: Unit.IUnit, underworld: Underworld, prediction: boolean) {
+  let units = prediction ? underworld.unitsPrediction : underworld.units;
+  // Find a random enemy unit and charm it
+  // Unit must be alive, in enemy faction, not a doodad, and not a miniboss
+  units = units.filter(u => u.alive && (u.faction != unit.faction) && (u.unitSubType != UnitSubType.DOODAD) && (!u.isMiniboss));
+  if (units.length > 0) {
+    const random = seedrandom(`${getUniqueSeedString(underworld)} - ${unit.id}`);
+    const chosenUnit = chooseOneOfSeeded(units, random);
+    if (chosenUnit) {
+      Unit.changeFaction(chosenUnit, unit.faction);
```
jdoleary, line 43: We should add heart particles for this.  I'll make a sprite for it

jdoleary (commented):
Fantastic work!! Can't wait to have these in the game!

src/cards/slow.ts
SoulMuncher: Slows are multiplicative, so applying 3 50% slows and applying a 150% slow do different things

Heavy toxins is written this way so that the rune can stack, without giving players the ability to apply slows at or over 100%.

We can't do it in the slow function either, because applying 20 1% slows is different than applying a 20% slow

50% slow + 50% slow = 75% slow
90% slow + 50% slow = 95% slow

src/modifierGolemancer.ts
SoulMuncher: That would be good!

src/modifierGolemancer.ts
SoulMuncher: Yes. If you have golemancer 5, and start your turn with 2 golems, it'll summon 3 more. If you start your turn with 5+, it wont summon any

src/modifierGolemancer.ts
SoulMuncher: I think I had issues with these functions but I didn't want to change them since they were being used in a lot of places already.

I'll look at it again and see what I can do about cleaning up the duplicate code

src/modifierBountyRefresh.ts
SoulMuncher: That might work... Is the unit "dead" when onKill events fire? I'll look into it

src/modifierAlchemist.ts
SoulMuncher: Similar discussion https://github.com/jdoleary/Spellmasons/pull/957#discussion_r1713032756

src/modifierCreepingDeath.ts
SoulMuncher: This was initially an onTurnStart effect and it felt like bad rng when it would reapply suffocate to things that were already suffocating, but now that it's onKill and it can chain... that doesn't necessarily sound like a bad thing... Yeah, we could have it add to existing stacks. It would feel better in situations where the whole map is already suffocated too, I'll update this

src/modifierGolemancer.ts
jdoleary: If you have a particular reason for not wanting to use it you don't have to.  I just wasn't sure if you were aware of it.  If you want to go ahead as is that's fine with me. 
Did the issues you have also stand as issues for the summoner? Maybe we should just refactor a single (or few) spawn functions in a util file?)

src/modifierGolemancer.ts
```diff
@@ -13,7 +13,7 @@ import { UnitType } from "./types/commonTypes";
 import { makeRisingParticles } from "./graphics/ParticleCollection";
 import { BLOOD_GOLEM_ID } from "./entity/units/bloodGolem";
 
-// Summon up to [quantity] golems each turn
+// Summon up to [quantity] golems each turn and grants the player summon golem spell
```
jdoleary, line 16: I still feel like this would work better as "summons X golems at the start of each turn" rather than topping off the player's number of ally golems.  It just feels confusing to me.
If you're dead-set on keeping it as is, we need to find a way to change the language so it's crystal clear

jdoleary (approved):

src/modifierGolemancer.ts
SoulMuncher: Yeahh.. it might be a bit confusing. I'm not super dead set on it, but here was my reasoning in case you were curious:

One issue I have with the summon playstyle is that it's 100% centered around hitting a "critical mass" of units that can finish the level on their own. You sit in a corner and repeatedly summon things until your army is bigger than their army, and that never felt very interesting to me. 

The vision is that Golemancer gives you a handful of summons to work with at any given time, without letting you reach that critical mass. The value of each summon is immediate: they block an attack, deal some damage, serve as a bloat bomb, or turn into a corpse for later use, and you almost want to use them up quickly to achieve whatever your current goal is. The cap ensures they'll never win the round on their own, forcing you to continue strategizing and adapting to the moment, and their immediate availability encourages interaction. You can never sit back and spam "end turn" and you're not just creating a wall or critical mass of summons

If I make it spawn X each turn and remove the cap, the rune is obviously much stronger, and we'll have to drastically increase the cost to balance it. It'll make the starting squad smaller and the end squad bigger, leaning more into the setup -> tipping point playstyle. Which is fine, just not what I was initially going for

The main goal was to diversify summon builds and create some variant of necromancer that calls for a little more strategy / creates more interesting scenarios.

.

Language wise, I'm not really sure...
- Summons X respawning golems
- Summons golems each turn, up to a limit of X.
- Creates a squad of X golems. Squad golems respawn each turn
- Summons X golems each turn. Cannot summon more than X golems at once

SoulMuncher: I'm happy for it to be implemented either way. There's definitely some value in making it more intuitive and consistent with other summon effects

src/modifierGolemancer.ts
SoulMuncher: I couldn't use the other functions for Alchemist because I wanted to spawn the potions a specific distance away from the player. Close enough they can grab it, but not so close its immediately picked up if players don't want to pick it up. I also didn't want the placement to be predetermined on a hexgrid because it's an awkward mechanic to play around. I wanted to use a custom radius for Golemancer, and it felt similarly cluttered/awkward with the hex grid imo.

FindValidSpawn seems a bit buggy too, where it sometimes doesn't spawn things? I'm not sure. We got a report about that / last will. There may have been some other logic I wanted to skip from the summoner function, but I haven't looked at it in a couple days now.

I do like the idea of refactoring into one or a couple generic functions in a util file, that would be much cleaner than how I wrote it

jdoleary: Okay this works! Thanks for the extra context

src/modifierGolemancer.ts
jdoleary: Oh I see.  I like that logic.  Let's go with:
"Creates a squad of X golems. Dead squad golems respawn each turn".

However, how does this handle if you use Summon Golem independent of the "squad"?

SoulMuncher: I'll need to update the logic to differentiate squad and non-squad golems if we go with that description.
Right now *any ally golem* counts towards the max, so its more like

"Each turn, if you have less than X golems, summon up to X golems"
"Summon golems each turn. Minimum of X summoned golems at the start of each turn"
"Golems are automatically summoned each turn if you have less than X golems"

jdoleary: Or more simply:
"Creates enough ally golems to ensure you have X at the start of each turn"

SoulMuncher: Yeah, that works!

PR #1405, opened 2025-08-12
Update cardUtils.ts

Gamer-Guy123:
Added support for the new proposed onCostCalculation event

### DIFF
--- a/src/cards/cardUtils.ts
+++ b/src/cards/cardUtils.ts
@@ -27,6 +27,7 @@ import { precisionId } from "../modifierPrecision";
 import * as Cards from "../cards";
 import { IUnit } from "../entity/Unit";
 import { fairIsFairId, soulmuncherId, witchyVibesId } from "../modifierDeathmasonConstants";
+import Events from "../Events";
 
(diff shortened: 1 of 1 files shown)


src/cards/cardUtils.ts
```diff
@@ -257,7 +257,13 @@ export function calculateCostForSingleCard(card: ICard, timesUsedSoFar: number =
         if (card.category == CardCategory.Targeting && caster.unit.modifiers[affinityTargeting]) {
             cardCost.manaCost = Math.floor(cardCost.manaCost * (1 - (0.01 * caster.unit.modifiers[affinityTargeting].quantity)));
         }
-
+        const events = [...unit.events];
+        for (let eventName of events) {
+            const fn = Events.onCostCalculationSource[eventName];
+            if (fn && caster) {
+              fn(caster, timesUsedSoFar);
```
jdoleary, line 264: Since I assume this is intended to alter the card cost it should probably return a CardCost and the fn should override the existing card cost like
`cardCost = fn(caster, timesUsedSoFar)`

